PR #14, opened 2017-09-16
Annealed Copper Bolts #1973

MineAnPlay:
Yes I know it's not realistic but we need it in game.


Dream-Master:
Let ne so this a different Way. There is a section ore processing where the recipe need to be added

Dream-Master:
Allready added

PR #27, opened 2018-04-10
General Improvments

bartimaeusnek:
+ Added GT_Recipe_Remover Class and moved all removeRecipeByOutput from GT_CraftingRecipeLoader class to it
+ Pulled Alloysmelter recipes from CoreMod minetweaker script to the GT_MachineRecipeLoader
+ Fixed broken AlloySmelter Recipes for ItemCasings (take in 3 instead of 2 and give out 2 instead of 3, titanium uses plates instead of ingots)
+ added methodes to get the coremods blocks as ItemStacks
+ added a method to get the coremods items with a custom amount as stack
+ enhanched perfomance/ram usage a bit (by using final, byte instead of int and ++i instead of i++ in for loops)

CoreMod.zs needs to be edited, I could outcomment or delete the AlloySmelter Recipes


src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
```diff
@@ -1460,14 +1460,38 @@ private void run2() {
 
         // --- Stargates
         if (Loader.isModLoaded("GalaxySpace") && Loader.isModLoaded("SGCraft"))
-            for (int i = 0; i < 6; i++)
+            for (byte i = 0; i < 6; ++i)
```
bartimaeusnek: i++ :
    create a temporary copy of i
    increment i
    return the temporary copy

++i :
    increment i
    return i

src/main/java/com/dreammaster/gthandler/GT_Recipe_Remover.java
```diff
@@ -0,0 +1,185 @@
...
+        GT_ModHandler.removeRecipeByOutput(GT_ModHandler.getModItem("opensecurity", "opensecurity.rfidReaderCard", 1L, 0), true, false, true);
+        GT_ModHandler.removeRecipeByOutput(GT_ModHandler.getModItem("opensecurity", "opensecurity.secureNetworkCard", 1L, 0), true, false, true);
+        GT_ModHandler.removeRecipeByOutput(GT_ModHandler.getModItem("opensecurity", "opensecurity.securityDoor", 1L, 0), true, false, true);
+        GT_ModHandler.removeRecipeByOutput(GT_ModHandler.getModItem("opensecurity", "opensecurity.securityDoorPrivate", 1L, 0), true, false, true);
+        GT_ModHandler.removeRecipeByOutput(GT_ModHandler.getModItem("opensecurity", "opensecurity.damageUpgrade", 1L, 0), true, false, true);
+        GT_ModHandler.removeRecipeByOutput(GT_ModHandler.getModItem("opensecurity", "opensecurity.cooldownUpgrade", 1L, 0), true, false, true);
+        GT_ModHandler.removeRecipeByOutput(GT_ModHandler.getModItem("opensecurity", "opensecurity.energyUpgrade", 1L, 0), true, false, true);
+        GT_ModHandler.removeRecipeByOutput(GT_ModHandler.getModItem("opensecurity", "opensecurity.movementUpgrade", 1L, 0), true, false, true);
+        //Computronics
```
Dream-Master: Did the change make computronics recipe remove working @bartimaeusnek ?

draknyte1:
> enhanched perfomance/ram usage a bit (by using final, byte instead of int and ++i instead of i++ in for loops)


Probably pointless optimisations. JVM probably does either or both already. 

I++ vs ++I have totally different use factors. As seen by your change, that wouldn't work unless I becomes -1.
I++ will give you the value of I, then increment the variable. 
++I will increment the variable and then return it.

src/main/java/com/dreammaster/gthandler/GT_Recipe_Remover.java
bartimaeusnek: havent touched computronics here @Dream-Master

src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
codewarrior0: Bytecode generated is identical in both cases!

bartimaeusnek:
@draknyte1 in for-loops prefix and postfix do the same thing. This might be pointless in terms of performance gain, but it forces the compiler to use the faster implementation of incementaion here.

src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
bartimaeusnek: Then this is pointless.

draknyte1:
No, because you'll get a different result for Object[i] if you use ++i vs i++ in a for loop. Hence why you'd have to start the index at -1, not 0, because ++i will skip 0.

bartimaeusnek:
The way for loop is processed is as follows

1 First, initialization is performed (i=0)
2 the check is performed (i < n)
3 the code in the loop is executed.
4 the value is incremented (++0 = write 0+1, 0++ = read 0, write +1, in theory. in most cases the compiler will use ++i but sometimes it does not get it right.)
5 Repeat steps 2 - 4

This is the reason why, there is no difference between i++ and ++i in the for loop which has been used.

Namikon:
I love such kind of conversations where you dissect the bytecode of something simple like this. It pulls your attention to stuff that can be learned where you didn't expect it 😁 
So far looks fine; The getItemStack methods for Blocks and Items probably should go in YAMCore. However, the ItemDescriptor class there is/was supposed to allow easy conversion between Items, Itemstacks with and without NBT data. But doesn't hurt to have these things here 👍

PR #32, opened 2018-04-21
Modified Galactic Greg def file

bartimaeusnek:
+ modified Europa OreGen to generate in the Ice Layer instead ontop of it
+ enabled TcetiE, CentauriA, Europa gen
+ added final modifiers for clear programming / potential memory reduction


draknyte1:
> added final modifiers for clear programming / potential memory reduction

We've established that JavaC does this automatically during compilation, so it's pointless to finalise every variable, method and object. Hell, unless it's a string or basic type, the final keyword does nothing at all.

bartimaeusnek:
@draknyte1 It still shows that this Variable wont be overwritten anymore, therefore its clearer programming.
see:
Java Language Specification
4.12.4. final Variables
A variable can be declared final. A final variable may only be assigned to once. Declaring a variable final can serve as useful documentation that its value will not change and can help avoid programming errors.

draknyte1:
Final variables aren't final whatsoever unless JavaC unlined it during compilation, which is actually my point. 

Private final Map x = new HashMap(); 

x is final, but remove that final modifier and I can change that final value. Be it public, private, protected or static..

PR #35, opened 2018-04-28
enabled Bedrock-Dim for OreGeneration via Galactic Greg

bartimaeusnek:
This enables the option to generate Veins in Bedrock dim.

### DIFF
--- a/src/main/java/com/dreammaster/galacticgreg/SpaceDimRegisterer.java
+++ b/src/main/java/com/dreammaster/galacticgreg/SpaceDimRegisterer.java
@@ -76,6 +76,7 @@ public class SpaceDimRegisterer
         registerModContainer(Setup_GalactiCraftCore());
         registerModContainer(Setup_GalactiCraftPlanets());
         registerModContainer(Setup_GalaxySpace());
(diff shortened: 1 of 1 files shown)


bartimaeusnek:
also fixes https://github.com/GTNewHorizons/GT5-Unofficial/issues/44

Namikon:
Have you verified that it generates ores? GalacticGreg was written to generate ores in GalaxySpace Dimensions; I never actually tried to generate anywhere else

Dream-Master:
we need the ores generated like bedrock dim like not gt like

Namikon:
If it's the wish of @Dream-Master to generate GregTech Ores in any Dimension, with any choosable Generator/feature pattern, that should be a separate mod then. (Someone totally should start an RFTools/Mystcraft clone in GregTech 🤔)
Anyways, GalacticGreg was never meant to do anything else besides handling SpaceDimensions which are maintained by GalactiCraft. This is not a simple "addon"; In order to do that properly (Keeping the syntax/API style as it is) a lot of work is required

Dream-Master:
ok we need to fix then only the issue that gt ores can spawn in bedrock dim with thaumic tinker generator and no gt 00 ores

PR #37, opened 2018-05-05
Ethanol production with forestry fermenter costs too much energy

DarkShadow44:
This addresses https://github.com/GTNewHorizons/NewHorizons/issues/2890, with this change biomass production is now energy positive and can be used to create electricity, very much like biogas.

Since we can't make recipes for the forestry fermenter that have a different time or energy usage, we have to take all the fermenter recipes from forestry and add them into the gregtech brewing machine. Then we can adjust the ratio of the biomass produced and the ethanol from biomass. This makes the forestry 'Fermenter' and 'Still' obsolete though, since they are now worse than the gt machines.

We add all forestry fermenter fermenter recipes into gregtech brewery with the following changes:
- Turn conversion rate from "40L biomass to 12L Ethanol" to "40L biomass to 20L Ethanol"
- Double the biomass output
- Increase time for recipes: 0.4 seconds per 1L biomass
- Change energy usage for fermenter recipes: 24EU per 1L Biomass

Example calculation for potato (remember, with biogas 1 Potato is worth ~2460EU)
1 Potato + Water + 960EU -> 40L Biomass
40L Biomass + 384EU -> 20L Ethanol
20L Ethanol -> 3200EU (net gain1856EU, taking 16 seconds)

Example calculation for potato with juice (higher efficiency)
1 Potato + Water + 1440EU -> 60L Biomass
60L Biomass + 576EU -> 30L Ethanol
30L Ethanol -> 4800EU (net gain 2784EU, taking 24 seconds)

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
+++ b/src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
@@ -651,7 +651,7 @@ public class GT_MachineRecipeLoader implements Runnable{
                     }
                 }
 
-                GT_Values.RA.addDistilleryRecipe(ItemList.Circuit_Integrated.getWithDamage(0L, 1L, new Object[0]), Materials.Biomass.getFluid(40L), Materials.Ethanol.getFluid(20L), 16, 24, false);
+                GT_Values.RA.addDistilleryRecipe(GT_Utility.getIntegratedCircuit(1), Materials.Biomass.getFluid(40L), Materials.Ethanol.getFluid(20L), 16, 24, false);
             }
             catch(Exception e)
             {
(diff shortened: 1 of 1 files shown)


bartimaeusnek:
@DarkShadow44 have you checked if any of the recipes are based on Minetweaker scripts?

DarkShadow44:
Yeah, I looked through the scripts, but there was nothing.

bartimaeusnek:
oh and @DarkShadow44 this might cause problems with GregTech if the Recipes are registered there already, did you checked that aswell?

DarkShadow44:
Well, gregtech adds it's own "biomass to ethanol" recipe with the old "40 to 12" ratio - but NewHorizonsCoreMod comes first. Gregtech has inbuilt safety for that, so the recipe will not be added again - the second recipe is simply dropped silenty.

Regarding the fermenting recipes - gregtech doesn't add those. However, since an item might already have a conflicting recipe in the brewery, I check that with "alreadyHasRecipe".

PR #41, opened 2018-05-26
new recipe for #2999

bartimaeusnek:
new recipe for https://github.com/GTNewHorizons/NewHorizons/issues/2999
@noobyaran9


noobyaran9:
Mind explaining the recipe ? haha
i know its something like n amount of biochaff + 750mb of water turns into m amount of mulch in chemical bath. Whats the power cost and whats the amount?

bartimaeusnek:
1BioChaff + 750L water -> 8Mulch (100%), 4Mulch (33%), 4Mulch (25%), 500ticks (25Sec) at 30EU/t

noobyaran9:
make it 16 biochaff and leave only the 8 mulch sounds fine.

bartimaeusnek:
i thought about that, but the fermernter from forestry uses 1 wheat to generate 1 mulch... so 1 chaff (8wheat) should be at least 8mulch imo.

noobyaran9:
i guess 4 is alright.

Anarack:
If bulk production of mulch is allowed then mulch->biomass in the pyro oven needs to be nerfed. not sure why it existed in the first place. As this would allow more than 8.6x the bio mass, which can be fermented into fermented bio mass, produced from one bio chaff.

noobyaran9:
OH! thats bad, I guess that recipe need to be nerfed then. And mulch into biomass? What?

Anarack:
Yep 4 mulch 4000 water to 5000 biomass at 45 seconds at 10 eu/t...

PR #46, opened 2018-07-22
Fix to support disabling MACHINE_METAL tinting on machines.

richardhendricks:
Because of java/gregtech/forge mod loading order, have to implement the config option to disable MACHINE_METAL tinting here in the core mod.


richardhendricks:
It's a relic from the testing I had to do to determine the root cause of the colors not changing.  The actual code generated is probably identical with any reasonable level of optimization on the compiler's part.

PR #47, opened 2018-08-06
Comb Time and Voltage Adjustments

MineAnPlay:
Decreased all Comb processing times to 128 ticks (6.4s).
Made all voltages consistent (15/16 of tier voltage).
Changed Voltage of Chrome and Platinum to HV (both are accessible before HV so doesn't require EV/IV).
Increased the output of Lithium (50%), Salt (Full dust at 100%), Electrotine (50%), Thorium (50%), Meteoric Iron (50%), Desh (25%).
Disabled Cosmic Neutronium, Infinity Catalyst, Infinity because not worth or not needed.


Dream-Master:
128 ticks is a bit short in my eyes. Will be add the whole bee stuff to gt code very soon. Will make gendustry cfg generated bees deprecated.

MineAnPlay:
Because of the small amounts of items the combs give, making the recipes take longer makes most resources too expensive (Such as Thorium, Uranium and Naquadah are power generating materials but because of processing time they be _very_ energy negative that it's not even possible to obtain power from them). Instead it should be limited by the speed of your bees and the voltage you have.

Dream-Master:
ok but between 128 and 1200 ticks are ages :)

PR #48, opened 2018-08-31
New UV Circuits

bartimaeusnek:
re-added PicoCircuit and QuantumCircuit as UV and UHV Circuits
Nano Circuit is now registered as Neutronium Circuit


Dream-Master:
Will add it and merge it with my changes and additions i made in gt branch

PR #55, opened 2018-10-05
added circuits for quantum suit recipe

DualBallers:
Possibility of using all LuV circuits (except for nano mainframe) in the Quantum Suit crafting.


Dream-Master:
lets see if it works

PR #65, opened 2019-02-14
Update readme twitch desc

shawnbyday:
Update to README to add brief description of mod so we can upload to twitch. If they require more detail ping me and I'll add more.


Dream-Master:
@shawnbyday thanks

PR #74, opened 2019-05-19
re-configured Quantium Voltage to IV

bartimaeusnek:
+removed unnescarry ,new Object[0] for varargs methods


0lafe:
It's more an LuV thing being used for the T5 and later shits. LuV voltage is fine, even zpm was ok.  Issue is more with temps past 7200k, which is sort of the LuV cap. You don't get it until LuV post assembler/CA. It's post OsIr, an LuV smelt

PR #76, opened 2019-05-25
added Abandoned GregTechPlusPlus Aspects

bartimaeusnek:
+will hopefully doesnt cause corruption when loaded up


bartimaeusnek:
directly merged so we keep the time frame and doom can test if this prevents corruption

PR #78, opened 2019-07-14
BioLabChanges

bartimaeusnek:
+buffed Radox (first step 1 -> 10, second step 1 -> 100)
+implemented SeaWeed Recipe idea (with some changed numbers as follows:
- GelatinMixture produces 2000L instead of 1000L
- itemGelatin needs 2000L GelatinMixture instead of 1000L
- MeatRaw gives 125L instead of 36L
- UnknownNutrientAgar needs 1000L MeatRaw instead of 288L
- UnknownNutrientAgar produces 8000L instead of 1000L
- SeaweedBroth needs 8000L UnknownNutrientAgar instead of 1000L
- SeaweedBroth produces 50L instead of 100L
)

https://github.com/GTNewHorizons/NewHorizons/issues/4823


draknyte1:
Meat giving 36 was a majority if not absolute group choice by those who came up with this idea, why adjust it?

bartimaeusnek:
technical reasons, mostly due to the fact that an universal cell / fluid capsule / most other liquid containers hold 1000L instead of 144L, and i have just added these fluids as simple fluids and not as materials.

draknyte1:
> don’t use cells

Whether a material or a fluid alone, makes no difference. 
Resizable flasks exist and so do (I think I enabled them) the GT++ sized cells. (1, 16, 36, 144L), so using something different is a change from the ordinary of cells are really required. (Just let your multi accept 9 fluids in)

PR #81, opened 2019-08-24
strip it, strip it good

mitchej123:
image

image


mitchej123:
Addresses https://github.com/GTNewHorizons/NewHorizons/issues/4782

PR #82, opened 2019-08-30
added UHV & UEV Circuit Assembler

bartimaeusnek:
+added UIV & UMV Circuit Assembler (currently disabled)
+added UHV, UEV, UIV to the customLoader
+reworked comments on Loader Machines


src/main/java/com/dreammaster/gthandler/GT_CustomLoader.java
```diff
@@ -44,7 +44,31 @@
...
+                Materials.Bedrockium,
+                null),
+        UIV(null,
+                null,
+                null,
+                null,
+                null,
+                null,
+                null,
+                null);
```
Dream-Master: Guess we need to define here more materials in the future

PR #105, opened 2020-04-01
Fix world world accelerator power consumption

Glease:
If you configure the range to 1 in block mode, then switch to TE mode, HV WA consumes 1024EU/t. If you use the default, i.e. range max, it will consume 3072EU/t. The reduction in energy cost goes even higher for EV+ WA (max 85% reduction for ZPM WA, see screenshot below). The behavior persists after server restart. 

85% off!


Dream-Master:
There was some reson to code it like that.
@Namikon can you look into it. Thanks

Glease:
If you are refering to de2dfb2f33750b35cd8f79758e47eddaddc861bf and GTNewHorizons/NewHorizons#2088, then it's most likely an oversight. Anyway, I'll wait for his review.

Namikon:
Whoops, it is indeed a mistake. However I'd like to keep the *3 power consumption for non-TE mode. Can you update your PR?

Glease:
Fixed

Namikon:
Merged. Thanks for the fix

PR #107, opened 2020-04-19
fixed OreDict Names

bartimaeusnek:
+ they use unlocalised names instead of localised ones now
+ fixed ganymede typo


src/main/java/com/dreammaster/gthandler/GT_Loader_Items.java
```diff
@@ -483,7 +483,7 @@ private void registerItems()
 		CustomItemList.CeresStoneDust.set(ItemList.CeresStoneDust.getIS());
 		CustomItemList.EuropaIceDust.set(ItemList.EuropaIceDust.getIS());
 		CustomItemList.EuropaStoneDust.set(ItemList.EuropaStoneDust.getIS());
-		CustomItemList.GanymedStoneDust.set(ItemList.GanymedStoneDust.getIS());
+		CustomItemList.GanymedStoneDust.set(ItemList.GanymedeStoneDust.getIS());
```
AndrewAmmerlaan, line 486: Okay so I don't really know anything about java and all, so I might be wrong, but from my java-noob point of view it looks like you're replacing `GanymedStoneDust` with `GanymedeStoneDust`. In that case shouldn't the first `GanymedStoneDust` here also be renamed to `GanymedeStoneDust`?

Dream-Master: yes it was a mistake. I wrote it wrong and now it is been corrected. And yes all old items are missing.

bartimaeusnek: Thanks, didnt see that

PR #108, opened 2020-04-19
Advanced smd

Dream-Master:
Add Advaced SMD components as Piky suggest
GTNewHorizons/GT-New-Horizons-Modpack#5928


Dream-Master:
Will add more recipes later.

Dream-Master:
Not merge yet

Prometheus0000:
So these are basically worth 4 of the regular ones (but regular ones are 1:1 with the basic ones?).  You can start using them from the Good Integrated Circuit, used to make your first HV circuits. You HAVE to use them starting from Quantumprocessors (used in IV to make ZPM circuits) however, which is kind of insane. You barely start using the normal SMDs before they're thrown out the window. You also use PBI to make them so they're only available once you overclock your LCR to IV power. The ASMD Capacitor also uses HSS-S, which means you need both Iridium and Osmium, so no ZPM circuits until you're done the platinum stuff. Which also requires a DT at IV. The ASMD Resistor and Transistor recipes both look fine, but the SMD Diode uses IGP dust, so you NEED Indium to make the circuits. And future ones. They also don't seem to be used in all the higher tier circuits, probably because they were missed.

Personally, I think the intended place to use them was further along, Like the path to UV or UHV circuits. Just the Indium is really bad, but I'd like to hear what other people think. You do get 64 at once, and only use 4 per ZPM circuit though.

Dream-Master:
@Prometheus0000 is your suggestions just let the normal smd componenets in Luv circuits and up ?

ScriptedPiky:
HSS-S could be reverted to HSS-E instead. However, you do *optionally* need Indium for the Quantum tier circuits anyhow. Regular SMD components could be used for the Quantum tier circuits instead of the A-SMD's being required.

Prometheus0000:
Does anyone actually use Indium instead of Radon for Qbits? I do think we should be able to use the normal ones for the Quantum, and _maybe_ Crystal tiers. But I don't really have any experience with the Crystal tier. And would these actually be considered 'cheaper' to use with the lower tier circuits? I'd probably continue using the normal ones at a 4:1 ratio.

And remember replicating Indium is getting harder too.

GTNH-Colen:
I'd rather keep these PBI guzzling monsters away from crystals honestly, wetware and above seems more suited for this territory.

0lafe:
Maybe not quantum's but crystals seem perfect to start PBI usage. If anything quantum's fit better but pushing it to wetware+ seems too far. PBI is very much IV, and crystals are way past that

ScriptedPiky:
It's basically 500 Radon tradeoff vs 1/3rd Indium for the Wafers. Although, since 100L of Radon is essentially 1 tiny Uranium 238 dust and 1000L of Air, I can see why people would use the Radon usage over Indium (the Radon complexity is a different issue in itself). I also do see that this could get nerfed indirectly by the replicator nerf, and how public servers may lack the ores necessary for more Indium production, I now see that this should be around Wetwares and maybe even Crystal Circuits, seeing how many nerfs can absolutely devastate the balance.

Also, the A-SMD or regular SMD Diodes are only used for a few amounts of circuit recipes surprisingly. The only ones are the lower tier beginning circuits (LV~EV), then Elite Nanocomputer (IV), Master Quantum Computer (LuV), Wetware Supercomputer (UV), Wetware Mainframe (UHV), and some other various high tier circuits along with all of the Magneto Resonatic Circuits.

GTNH-Colen:
> Maybe not quantum's but crystals seem perfect to start PBI usage. If anything quantum's fit better but pushing it to wetware+ seems too far. PBI is very much IV, and crystals are way past that

Is it though? I didn't make a setup for PBI until like ZPM when I needed it for hulls and wetwares? I never considered it an IV tiered chemical.

Prometheus0000:
Personally I have no problem with PBI usage, I made a dozen stacks as soon as I could. It's great! I want to use it more than just for busses/hatches.

botn365:
i find crystals to be prety late and with C assline you unlock the soc version of it prety early
with quantum its a bit to early mostly ecous IV already has a lot of stuff defnetly with the platinum change

GTNH-Colen:
Yea I suppose actually considering the fact you get the non smd recipe for the IV-tier crystals at UV it's not actually that bad to use adv SMDs in that line.

Extremity4:
I think crystals are a good place to start using these so long as the component per circuit ratios are set such that the overall cost of Crystalprocessors doesn't spike massively.

Glease:
Quantum already requires loads of platinum, which implicitly require you to finish pt proc. This material gate is hard enough. No need to push it further. 

Crystals should allow normal SMD as well, but with a 2.5x more cost, i.e. 10 SMD/1ASMD with C assline or 60SMD/6ASMD with C assembler. This would make the transition a bit smoother, but place a great enough incentive to move to ASMD. If you think this is not enough, consider enforcing ASMD usage in crystal mainframes.

GTNH-Colen:
> Quantum already requires loads of platinum, which implicitly require you to finish pt proc. This material gate is hard enough. No need to push it further.
> 
> Crystals should allow normal SMD as well, but with a 2.5x more cost, i.e. 10 SMD/1ASMD with C assline or 60SMD/6ASMD with C assembler. This would make the transition a bit smoother, but place a great enough incentive to move to ASMD. If you think this is not enough, consider enforcing ASMD usage in crystal mainframes.

I like this option the most tbh, kinda satisfies most parties.

botn365:
i dont think crystal schould be able to use normal smds becous normals are already prety cheap and automating the pbi is a pain enough that people wil just use the normal ones anywhey

GeicowithaBango:
Since pbi is a IV plastic I would consider asmd parts for the crystal line. for proper understanding, the crystal line is unlocked after your first fusion reactor. 
therefore, the argument about a high cost of Indium is obsolete. furthermore, is the idea of an penalty for using normal smd parts at this point also laughable. 
the focus here should be to integrate new interesting stuff as ingredients for asmd. 
and frankly speaking, hss-e in its current state has barely any use. it would be an excellent alloy for the creation of asmd components.

GeicowithaBango:
Rubidium, cerium and higher tier gems could have nice usage too :)

0lafe:
Bring back the fine OsIr wire? ;D

PR #115, opened 2020-07-15
Added remaining Bacteria from Binnie's to GT Brewing machine

redmage17:
Confirm the recipes are working and crafted fluid is accepted by Binnie's Machines.


Dream-Master:
Will check recipe ingame

PR #126, opened 2020-10-23
Reuse warding focus when crafting warded glass

Glease:
Support for GTNewHorizons/GT-New-Horizons-Modpack#6846

This is a fix targeting a specific recipe. It can be expanded to allows ingredient reusing in any arcane recipes, but for now let's keep it simple.

### DIFF
--- a/src/main/java/com/dreammaster/modfixes/thaumcraft/ReuseWardFocusFix.java
+++ b/src/main/java/com/dreammaster/modfixes/thaumcraft/ReuseWardFocusFix.java
@@ -1,0 +1,57 @@
+package com.dreammaster.modfixes.thaumcraft;
+
+import com.dreammaster.modfixes.ModFixBase;
+import cpw.mods.fml.common.eventhandler.SubscribeEvent;
+import cpw.mods.fml.common.gameevent.PlayerEvent;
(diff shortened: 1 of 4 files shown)


Dream-Master (approved):

src/main/java/com/dreammaster/modfixes/thaumcraft/ReuseWardFocusFix.java
```diff
@@ -0,0 +1,57 @@
...
+        if (wardedGlassBlock == null)
+            return false;
+        wardedGlass = new ItemStack(wardedGlassBlock, 1, 2);
+        return (focusWard = GameRegistry.findItemStack("Thaumcraft", "FocusWarding", 2)) != null;
+    }
+
+    @SubscribeEvent
+    public void onItemCrafted(PlayerEvent.ItemCraftedEvent e) {
+        if (getIsActive()) {
+            if (e.craftMatrix instanceof TileArcaneWorkbench && wardedGlass.isItemEqual(e.crafting)) {
```
repo-alt, line 45: That is not enough for arcane crafting terminal and arcane assembler, perhaps just check if the crafting inventory is not null?

PR #130, opened 2020-11-11
Added ardite ingot to block recipe

Ethryan:
In the compressor

Should fix!: [#4602](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/4602) (Not tested ingame)


Prometheus0000:
? Then where's the original one that didn't use oredict? In GT5U?

Ethryan:
Might be a mt script, i will look for it.

Dream-Master (approved):

PR #136, opened 2020-11-16
Try to move TC metals to Core Mod

Ethryan:
(Not Tested) - did this during lunch at work


Dream-Master (changes requested):
Gt Brass and Tico Brass havent the same ingredients. Please rework it use only Tico brass to avaoid exchange between the two blocks.

Dream-Master (approved):

PR #144, opened 2020-12-09
add new recipe and add remove to code

botn365:


Dream-Master:
add two steel screws to the recipe or one screw and a screwdriver. and we need to move the quest as well.

Dream-Master:
I will add it now

PR #151, opened 2021-01-14
Moved Blocks to init phase

bartimaeusnek:


glowredman:
What purpose does this serve? Aren't blocks and items supposed to be loaded at preInit?

bartimaeusnek:
@glowredman they were @ postinit. they *should* exist in preinit, but these are GT Blocks, so putting them in preinit might cause issues, so i moved them to init instead.

glowredman (approved):

PR #152, opened 2021-01-18
Fix warding focus fix not fully working

Glease:
I have to use asm to patch warding focus. I'm well aware of the side effect of the addition of `getContainerItem()`. The previous method cause significant lag, and sometimes just don't work, if it does not use the `setInventoryContentSoftly` method. Some other method I tried will cause thaumenergistic's arcane assembler to throw concurrent modification exception and therefore won't work.

We could also just call the `setContainerItem()` in post init on `ItemFocusWarding`, but that method is not ItemStack aware, and it will wipe all the NBT data, e.g. wand enhancement (or whatever that was called) and custom names.

I'd be glad if anyone can find a better alternative. I don't really want to use ASM on these trivial things but this seems to be the only sane way so far.

Signed-off-by: Glease <4586901+Glease@users.noreply.github.com>


Prometheus0000:
We could also just change the recipe back to giving 64?

Glease:
I don't want to automate the production of warding focus. That'd be ridiculous.

repo-alt:
What code does that asm insert? I didn't see much asm to easily restore bytecode from it.

Glease:
The equivalent code in plain java is
```patch
package thaumcraft.common.items.wands.foci;
//import stuff
public class ItemFocusWarding {
    public ItemFocusWarding() {
-        this.setCreativeTab(/* something */);
+        this.setCreativeTab(/* something */).setContainerItem(this);
    }

+    @Override public ItemStack getContainerItem(ItemStack stack) {
+        ItemStack copy = stack.copy();
+        copy.stackSize = 1;
+        return copy;
+    }
// rest unchanged
}
```

Glease:
In retrospect there is little reason to do that stackSize = 1 reset, perhaps I should change it.

repo-alt (approved):

PR #155, opened 2021-02-22
Fix RawRubber being Rubber for slime centrifuging

Prometheus0000:
It seems that they were added with Rubber instead of RawRubber by accident in https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/143 and before that when moving from script to code. I left the purple block as rubber since you can't farm it.

If it was on purpose then please say so. You can apparently get them from fish traps and fish catchers, and you can certainly get them from the IC2 slimeplant or a mob farm, so it seems like this change would make sulfur a little useless if it was intended. As well as making rubber super easy to get.


Dream-Master:
no rubber is fine. compared to the green slime which gives rubber as well. If we change this we must change green rubber too. @Prometheus0000

Prometheus0000:
Line 1170 is about green slime already?

Dream-Master:
yea didnt see it on my mobile.

Dream-Master (approved):

PR #156, opened 2021-02-23
Fix classloading order when not all mods are loaded

Glease:
Fixes GTNewHorizons/GT-New-Horizons-Modpack#7515

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/GT_Loader_CasingNH.java
+++ b/src/main/java/com/dreammaster/gthandler/GT_Loader_CasingNH.java
@@ -1,0 +1,13 @@
+package com.dreammaster.gthandler;
+
+import com.dreammaster.gthandler.casings.GT_Block_CasingsNH;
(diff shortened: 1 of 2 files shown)


bartimaeusnek:
wat?

Dream-Master:
?

repo-alt:
it requires a Java dev to understand that

bartimaeusnek:
I guess, this encapusulates the GT_Block_CasingsNH class, so it doesnt get loaded during initialisation of the mod container. still a bit weird for a fix. Maybe we should move the ModContainer as a whole, or create a Pipeline (which Forge already offers) to prevent such weird breakages in the future.

Glease:
Somehow the verifier fires before the method is called. This ensures the verifier doesn't see classes not yet present on classpath.

repo-alt (approved):
This is a piece of black magic, but if it works, ok.

PR #157, opened 2021-02-24
Fix baklava giving fire resistance

Elisis:
See: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/7359


bartimaeusnek:
How about we just do this instead of this whole "detection" function? Should be sufficient...

```JAVA
ItemData assoc = GT_OreDictUnificator.getAssociation(pHeldItem);
if (assoc == null || assoc.Material == null || assoc.Material.Material != Materials.Lava)
    return false;
return true;
```

Dream-Master:
@Elisis

Elisis:
Yeah, I'll change it when I can.

Dream-Master:
@bartimaeusnek

bartimaeusnek:
Needs to be tested. Since I don't have any idea if it will really work out in-game.

Dream-Master:
I merge this now so we can test it on zeta later

Dream-Master (approved):

PR #158, opened 2021-03-03
Remove non-working and pointless honey changing recipes

Prometheus0000:
Use the centrifuge one instead, these ones didn't work anyway.

PS: the fermenter recipes for biomass use the wrong type of honey (BoP), but I'm not sure how to fix it.


Dream-Master:
i add this recipe ages ago to get the right honey. Maybe the original recipe was fixed (ore dict not work on fluids I guess)

Prometheus0000:
This is the recipe ingame
image
It's the same thing on both sides.
There's a one-way centrifuge recipe that works though
image

bartimaeusnek (approved):

PR #160, opened 2021-03-13
Fix alu gravel not giving chance to get bauxite 

Prometheus0000:
When it was moved to code, it was accidentally set to alu dust, when it should be bauxite when macerating. This fixes that.


Dream-Master:
are you get alu gravel when you use ender io grinding machine ?

Prometheus0000:
No idea. Well, assuming you mean dust. You never get gravel from a machine. I just looked at what you changed and fixed it.

Dream-Master (approved):

PR #172, opened 2021-03-30
Buff air filter multi's effectiveness

Glease:
According to boubou's calculation and feedback from discord folks, this new formula seems to give a satisfying number for HV+.

New formula: each muffler reduce `300 * displayed turbine efficiency * Floor(2.5 ^ min(multi tier, muffler tier)) / 10` gibbl every second
New Requirement: Power tier of machine MUST match the tier of muffler or the muffler would underclock.

Key indicators: 
* Standard HV equipments: ~4k gibbl/second 
* Standard EV equipments: 15k~20k gibbl/second
* UHV equipments: ~9m gibbl/second
* Turbine last twice as long as it would inside a turbine generating the same amount of EU as the air filter consumes.

Calculation: https://discord.com/channels/181078474394566657/502566698142924811/826237068845383691
Feedback: https://discord.com/channels/181078474394566657/502566698142924811/826422610807881728


Glease:
I promise this would be the last force push. The old turbine damage code seems wrong in the first place and it makes me really confused.

repo-alt:
Eh, looks like this version is exactly the same as what I was reviewing?

repo-alt (approved):

Glease:
Maybe it is, maybe it's not. I'm a bit overwhelmed by this piece of stupid code.

johnch18:
Will the filter only work on one chunk or is it an area?

PR #176, opened 2021-04-24
Make gravisuit armor hover mode not fall when idle

Glease:
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/7855


glowredman (approved):
Looks good

PR #189, opened 2021-05-26
Fix neutron reflectors

Prometheus0000:
See https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/7752, https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/7986.

Changed which ones are x5 and x8, since the ones with durability don't work right in the assembler.


Dream-Master:
will check

Prometheus0000:
Why did you merge one but not the other? It only works right with both merged.

Dream-Master:
because i need to check them. Just to tiered to compare lines or go ingame now. If i check them i will do.

Dream-Master (approved):

PR #191, opened 2021-06-14
Make output bus (ME) cheaper

Prometheus0000:
See https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/7972

Remember that the point is to get people to make them as soon as humanly possible to reduce lag, so making them expensive is counter to that goal. Making them cost 2-4x as much as a normal bus+interface only convinces people to _not_ use them.

I made it the HV output bus so you can reuse your old ones, BTW.


Dream-Master (changes requested):
Ae2 need titanium so hv make no sense or did somone change that without my knowing? I ask repo who have a lot of experience with ae and he suggested make it EV.
I not approve this. reducing components is another think we have to check.

Prometheus0000:
> Ae2 need titanium so hv make no sense or did somone change that without my knowing? I ask repo who have a lot of experience with ae and he suggested make it EV.

Like I said, I made it the HV output bus so you can reuse your old ones, otherwise you end up with a bunch of them leftover. Repo also said in the issue

> (not like I'm against making it cheaper, just a technical note)

Did you not read the linked issue?

> I not approve this. reducing components is another think we have to check.

What are you checking? Do you understand the point of making it cheaper?

Prometheus0000:
What you need to ask yourself is "why would a player make this with the current recipe?" The benefits of it go to the server owners, not the players. With the current recipe, there's no reason to make them unless you're UV/UHV and your SP game is slowing down. But my understanding is that most people don't get there. But server owners get the benefit immediately, as soon as you start using it. Honestly they should be _cheaper_ than the current output bus+interface combo, but I figured you wouldn't allow that.

Dream-Master:
> What you need to ask yourself is "why would a player make this with the current recipe?" The benefits of it go to the server owners, not the players. With the current recipe, there's no reason to make them unless you're UV/UHV and your SP game is slowing down. But my understanding is that most people don't get there. But server owners get the benefit immediately, as soon as you start using it. Honestly they should be _cheaper_ than the current output bus+interface combo, but I figured you wouldn't allow that.

Your constant complaining that supposedly something is too expensive annoys me a bit. I'll take a look at the ingame and check recipe later.

Prometheus0000:
I just thought that rephrasing what I said might be better way to help you understand.

Dream-Master:
let me check that recipe ingame because idk what components are used out of my head when see "appliedenergistics2", "item.ItemMultiPart", 2L, 440

Prometheus0000:
440=ME interface, 30=acceleration card (this uses titanium BTW).

repo-alt (approved):
I don't think it really matters (you can disassemble the hatch anyway), but I have no objections either. Yes, this device benefits the server, not the player, but higher tier players usually do care about their base performance.

Prometheus0000:
I initially missed your point

> Indeed ME output bus is more CPU friendly even before that, because it involves no slot handling, but, before you start counting in stacks per tick the effect is minimal.

and figured that it would have a real effect right away.

repo-alt:
It will, just the effect will become noticeable only when the throughput is high enough.

Dream-Master (approved):

PR #193, opened 2021-06-15
Add isloaded checks and organize/format file

Prometheus0000:
I tried to put it mostly in order of GT, coremod, crossmod recipes.

BTW, the teleporter recipe was removed because it's not even used, not to mention the current one (script I think) is uncraftable anyway, so I dunno why we even let it have a recipe.

I dunno what's up with the ```Object[] o = new Object[0];``` line, but I'm just going to leave it alone just in case, since I don't feel like messing with it.

There should be no functional change.


Dream-Master (approved):
Need to check ingame but if no line is missing all fine

PR #197, opened 2021-07-27
Add helper to create pyrolyse oven recipes for logWood

D-Cysteine:
Add helper class to create pyrolyse oven recipes for `logWood`. Where this class is currently registered to run (GT post-load) works; I know that GT++ postinit will also work. Earlier than that, some woods may not have been registered in the ore dictionary yet. Let me know if there is a better place to put this code.

I've tested that with this change, all `logWood` woods work in the pyrolyse oven. This will allow us to delete the BW and GT dynamic pyrolyse recipe generation code which is causing lag.

I deleted the `ExtraTrees` pyrolyse recipes in `GT_MachineRecipeLoader` because they are just a copy of the method the new code is calling:
https://github.com/GTNewHorizons/GT5-Unofficial/blob/experimental/src/main/java/gregtech/loaders/oreprocessing/ProcessingLog.java#L97-L107
so that code is now redundant.
 * BW code removed in https://github.com/GTNewHorizons/bartworks/pull/19
 * GT code removed in https://github.com/GTNewHorizons/GT5-Unofficial/pull/608
 * See: GTNewHorizons/GT-New-Horizons-Modpack#8047

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -1,6 +1,6 @@
 minecraft.version=1.7.10
 forge.version=10.13.4.1614
-coremod.version=1.7.10-1.8.01
+coremod.version=1.7.10-1.8.02
 
 
 yamcore.version=0.5.78
--- a/src/main/java/com/dreammaster/bartworksHandler/PyrolyseOvenLoader.java
+++ b/src/main/java/com/dreammaster/bartworksHandler/PyrolyseOvenLoader.java
@@ -1,0 +1,24 @@
+package com.dreammaster.bartworksHandler;
+
+import gregtech.loaders.oreprocessing.ProcessingLog;
+import net.minecraft.item.ItemStack;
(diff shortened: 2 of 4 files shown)


Glease:
Did you test if it works? I'm not totally sure if GT after post load is late enough to have all logWood entries.

D-Cysteine:
Yup, went in-game, did an NEI search for `#logWood`, and checked that they all worked (though admittedly I only checked a few woods for like Extra Trees and Forestry which have a ton). I also checked that the actual list of things returned by `OreDictionary.getOres("logWood")` is equal in-game and during GT postload.

One thing that I'm not sure of, though, is if the mod loading order is stable. Is it a concern that the order might randomly change and this would stop working?

Glease (approved):

PR #202, opened 2021-08-20
fix disassemble dupe

GlodBlock:
fix https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/8426


Prometheus0000:
So you just made them not disassemble-able? I don't think that's a proper solution.

GlodBlock:
no, i only make the crafting table recipe can be disassembled. it is still disassembleable for its assemble recipe.
图片

Dream-Master:
i guess there houndet of recipes have this tag. Not sure if we should remove this tag from whole gtu5 ?

GlodBlock:
not necessary. if all of these tags are removed, the disassembler won't be able to disassemble any matchines or something that doesn't have a assemble recipe.

Dream-Master (approved):

PR #206, opened 2021-09-06
Add Glue Balls to Board recipes

Dream-Master:
* Use Glue Balls instead of sticky risin when make Basic Circuit boards
* Add control circuits to all assembler recipes
* Add Circuit part recipes with slime balls/glue


Prometheus0000:
Will people actually use this though? Getting glue from the smeltery is a huge pain after all.

Dream-Master:
Yea maybe i need to think off an alternative but i dont like it be only sticky resin. Maybe rubber dust ?

Prometheus0000:
What about adding an oredicted version of slimeballs (might be a different oredict, I can't see it on the stripped pack I have)? Like the recipe for magma cream. That would mean glue, slime balls of both types, blood drops all would be usable.

You also need to keep in mind that after you get the assembler you're just going to skip making coated boards and make circuit boards directly. So making it too advanced (I thought of adding an assembler GT glue recipe) won't work out either.

Dream-Master:
> What about adding an oredicted version of slimeballs (might be a different oredict, I can't see it on the stripped pack I have)? Like the recipe for magma cream. That would mean glue, slime balls of both types, blood drops all would be usable.
> 
> You also need to keep in mind that after you get the assembler you're just going to skip making coated boards and make circuit boards directly. So making it too advanced (I thought of adding an assembler GT glue recipe) won't work out either.

Yes the slimeball idea ordict comes in my mind but i dont find any existing one. Maybe i need to create one.
Sure it is only before assembler. I am atm in this situation. Have a bender working on a Lathe and after this assembler.

Dream-Master:
@Prometheus0000 add more recipe variants

Prometheus0000:
OK

Dream-Master:
image
image

PR #210, opened 2021-09-28
Move Ztones from script to coremod

Prometheus0000:
Also add assembler recipes for the blocks
Also resort the RR file alphabetically and use vars for repeated strings
Also changed to delayed version for the removed recipes. All the recipes in the crafting file are buffered, so it should be ok?

This should fix https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/7736
The ztones script file needs to be removed after merging.


Dream-Master (approved):

Dream-Master:
@Prometheus0000 guess you tested it ingame

Prometheus0000:
Yup

PR #214, opened 2021-10-21
balance the NAND recipe

GlodBlock:
using this recipe:
image

https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/8715


GlodBlock:
> Will this add a CAL recipe automatically as it has circuit in the name? Or is a CAL recipe intended in any way?

it will auto add CAL recipe if it has 'circuit' in its oredict name or unloaclized name.



> Why did you make a PR? Isn't it fine the way it is? Aren't they excessively cheap now?

GTNewHorizons/GT-New-Horizons-Modpack#8715

GlodBlock:
how many output do you think is reasonable

Prometheus0000:
Well first off, is a nand chip actually cheaper than vacuum tubes? I didn't see any comparisons.

Not to mention that you can't use nand chips in making circuits (and you shouldn't), so their primary use is in tanks (which become obsolete) or AE drives (was there any others?), meaning there's highly limited usage of them in the first place.

GlodBlock:
nand needs higher tech to produce so it should be cheaper. and as you said, nand has more limited usage than vacuum tube. if it is not cheaper, nobody will make it.

PR #230, opened 2021-11-16
one step mix recipe for some eio alloy

GlodBlock:
add one step eio alloy mixer recipe:
 - Conductive Iron
 - Energetic Alloy
 - Energetic Silver
 - Vivid Alloy


iouter:
why reduce time? it takes 5s to mix each time, so it corresponds 8eu/t 20s, 65s, 200s.

Dream-Master:
i would say make the recipe mv/hv otherwise the other recipes are obsolete.
Process time and used eu need to be more than the old recipes per item

GlodBlock:
now i have changed them into mv/hv, and they cost more eu per item than normal recipe

Dream-Master (approved):

PR #237, opened 2021-12-02
Chisel

bombcar:
update build and add a recipe for a chisel block (factory) to the assembler


Ethryan (commented):
Stone and Iron are rather common materials.
Tink we needo some kind of circuit here.

bombcar:
> Stone and Iron are rather common materials.
> I think we need some kind of circuit here.

I can pull it up but any idea of a good “chisel” circuit I could use?

Ethryan:
Hmm, 15 maybe?

leagris:
I think it is a bad idea to spam the assembler recipes map with decoration crafts an d other mundane things like that.
I'd say re-implement and modernise the electric workbench from GT3-4 and have it handle the workbench recipes if you like Gregtech ability to craft these things.

bombcar:
> I think it is a bad idea to spam the assembler recipes map with decoration crafts an d other mundane things like that.
> I'd say re-implement and modernise the electric workbench from GT3-4 and have it handle the workbench recipes if you like Gregtech ability to craft these things.

An electric chisel-like would be nice, time to learn how to add a machine.

We had decent amount of lag traced back to translocators on an auto chisel.

bombcar:
@draknyte1 has dibs on a chisel multi block

PR #238, opened 2021-12-05
Add new electric air filters

boubou19:
This PR adds new electric air filters to the game:

Left one is the tier 1, availiable in LV, has a working area of 3x3 chunks.
Middle one is the tier 2, availiable late HV/early EV, has a working area of 5x5 chunks.
Right one is the tier 3, availiable in late EV/early IV, has a working area of 7x7 chunks.

The controller has 2 modes:
- a normal mode, where one chunk is processed at a time in the working area, in order.
- a random mode, where the chunk being cleaned is picked randomly between all the polluted chunks in the working area.
To cycle between the modes, right click on the controller with a screwdriver.
image

The controller can change its working area from 1x1 to (2\*tier+1) x (2\*tier+1) chunks. To do that, right click with a screwdriver on the controller while sneaking. The default size is the maximum working area.
image

The pollution removal calculation is equals to 30 \* Muffler Hatches \* Turbine Eff \* Multiblock Eff \* 2.5 ^ Effective Tier where:
- Muffler Hatches is the amount of muffler hatches on the multiblock (up to 8)
- Turbine Eff is the efficiency of the turbine chosen to run the multi
- Multiblock Eff is the efficiency of the multiblock (100% when no maintenance issue, -10% per issue)
- Effective Tier is the Effective Tier of the multi: the minimum tier between the electric tier of the multi and the tier of the muffler hatches

The pollution cleaning is boosted by two when using an Absorption Filter. When added in the input bus, it will be consumed and used for 30 successful pollution cleaning cycle (if a chunk with no pollution in it is processed, that doesn't count as successful). After the 30th successful pollution cleaning cycle, the output bus will give a Dirty Absorption Filter.
To know if an Absorption Filter is present and how many remaining cycles there are, use an portable scanner on the controller:
image

recipes:
- Filter Turbine Casings:
image
image

- Filter Vent Casings:
image
image

- Electric Air Filters:
image
image


src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
```diff
@@ -229,15 +229,36 @@ public void run() {
...
+        GT_Values.RA.addAssemblerRecipe(new ItemStack[]{ItemList.Hull_IV.get(1L), ItemList.Casing_Turbine3.get(1L), GT_OreDictUnificator.get(OrePrefixes.rotor, Materials.TungstenSteel, 1L), ItemList.Electric_Motor_IV.get(2L), ItemList.Electric_Pump_IV.get(1L), GT_OreDictUnificator.get(OrePrefixes.cableGt01, Materials.Tungsten, 1L), GT_Utility.getIntegratedCircuit(1)}, Materials.Tin.getFluid(288), CustomItemList.Machine_Multi_AirFilterT3.get(1L), 200, 120);
+        GT_Values.RA.addAssemblerRecipe(new ItemStack[]{ItemList.Hull_IV.get(1L), ItemList.Casing_Turbine3.get(1L), GT_OreDictUnificator.get(OrePrefixes.rotor, Materials.TungstenSteel, 1L), ItemList.Electric_Motor_IV.get(2L), ItemList.Electric_Pump_IV.get(1L), GT_OreDictUnificator.get(OrePrefixes.cableGt01, Materials.Tungsten, 1L), GT_Utility.getIntegratedCircuit(1)}, Materials.Lead.getFluid(576), CustomItemList.Machine_Multi_AirFilterT3.get(1L), 200, 120);
```
boubou19, line 260: If someone could tell me why the fluids aren't added in the recipe in game, because despite adding 3 recipes for each block in the assembler, i only get one with no fluid.

GlodBlock: you should use getMolten()

iouter:
I think the stage that needs to clean up the pollution most is mv-hv, but this stage can only use tier 1, it is better to change the titanium to stainless steel or aluminum.

PolkageistHS:
Currently the earliest you can use these is MV, because they require a small turbine that cannot be crafted until MV.  The GT++ iron/bronze turbines do not work.  

Also, when I ran this in my base, pollution consistently went up, not down, as I ran it.  The only thing I had running in the chunk was an LV combustion gen to power it, with a small blue steel turbine and 8 LV mufflers on the multiblock scrubber.  When I shut the multiblock off, pollution did not change, and when I turned it back on it started going back up by about 100 every second.

Dream-Master:
> Currently the earliest you can use these is MV, because they require a small turbine that cannot be crafted until MV. The GT++ iron/bronze turbines do not work.
> 
> Also, when I ran this in my base, pollution consistently went up, not down, as I ran it. The only thing I had running in the chunk was an LV combustion gen to power it, with a small blue steel turbine and 8 LV mufflers on the multiblock scrubber. When I shut the multiblock off, pollution did not change, and when I turned it back on it started going back up by about 100 every second.

So we have a air Polluter rn

src/main/java/com/dreammaster/gthandler/multiAirFilter/GT_MetaTileEntity_AirFilterBase.java
```diff
@@ -3,95 +3,135 @@
...
     protected int mPollutionReductionWholeCycle =0;
-    protected int mAccumulatedPollutionReduction =0;
-    protected int baseEff = 2500;
+    protected int baseEff = 0;
+    protected int multiTier = 0;
+    protected int chunkIndex = 0;
     protected boolean hasPollution=false;
-    static final GT_Recipe tRecipe= new GT_Recipe(
+    protected int[][] chunkList; //list of the chunks, each entry contains the x and z coords of the chunks
+    protected int mode = 0; // 0 for processing chunks in order, 1 for processing chunks randomly
```
TheElan, line 44: Better to use enum in such cases, more descriptive

TheElan, line 44: Better to use enum in such cases, more descriptive

src/main/java/com/dreammaster/gthandler/multiAirFilter/GT_MetaTileEntity_AirFilterBase.java
```diff
@@ -3,95 +3,135 @@
...
+public abstract class GT_MetaTileEntity_AirFilterBase extends GT_MetaTileEntity_MultiBlockBase {
     protected int mPollutionReductionWholeCycle =0;
-    protected int mAccumulatedPollutionReduction =0;
-    protected int baseEff = 2500;
+    protected int baseEff = 0;
+    protected int multiTier = 0;
+    protected int chunkIndex = 0;
     protected boolean hasPollution=false;
-    static final GT_Recipe tRecipe= new GT_Recipe(
+    protected int[][] chunkList; //list of the chunks, each entry contains the x and z coords of the chunks
```
TheElan, line 43: Can be named `chunkCoordinates`. Also having a two dim array complicates things, you might be able to use some concrete type here. Probably there is a Vector or Point class somwhere, which has xy fields. If there is none, I would still go with typed approach and create small custom class for holding this data and have it here like like `PollutionChunk[]`

src/main/java/com/dreammaster/gthandler/multiAirFilter/GT_MetaTileEntity_AirFilterBase.java
```diff
@@ -3,95 +3,135 @@
...
+                .addEnergyHatch("Any bottom layer casing")
+                .addMaintenanceHatch("Any bottom layer casing")
+                .addInputBus("Any bottom layer casing")
+                .addOutputBus("Any bottom layer casing")
+                .toolTipFinisher("GTNH Coremod");
+        if (!Keyboard.isKeyDown(Keyboard.KEY_LSHIFT)) {
+            return tt.getInformation();
+        } else {
+            return tt.getStructureInformation();
+        }
```
TheElan, line 114: Would be nice if builder had a method for optional/hidden/alternative info

src/main/java/com/dreammaster/gthandler/multiAirFilter/GT_MetaTileEntity_AirFilterBase.java
```diff
@@ -3,95 +3,135 @@
...
+
+    public int getCasingIndex(){
+        if (multiTier <=1){
+            return 57;
+        }
+        else if (multiTier == 2){
+            return 59;
+        }
+        else{
+            return 60;
```
TheElan, line 133: "magic numbers", would suggest extracting as constants to class which is responsible on tier description

src/main/java/com/dreammaster/gthandler/multiAirFilter/GT_MetaTileEntity_AirFilterBase.java
```diff
@@ -117,104 +157,144 @@ public boolean checkRecipe(ItemStack aStack){
         mEfficiencyIncrease = 10000;
         mEfficiency = 10000 - (getIdealStatus() - getRepairStatus()) * 1000;
         mPollutionReductionWholeCycle = 0;
-        mMaxProgresstime = 200;
-        mEUt=-32;
+        // check pollution for next cycle:
+        hasPollution = getTotalPollution()>=10000;
```
TheElan, line 161: What do you think about extracting values like these to a config? Probably would be nice to have other efficiencies and coefficients there too.

TheElan (commented):
Don't see any apparent reason for pollution to go up. I have three guesses:
1. Didn't notice original code calling pollution removal, maybe addPollution actually adds absolute value and you need to use other method for reduction
2. Somewhere you get minus in equation (shouldn't be an overflow, but who knows), try to check how it behaves without minus when adding pollution and to check the value in debugger mode with breakpoint on addPollution call
3. Pollution values are unbalanced and generator is able to produce more than scrubber scrubs (again would check with debugger/logs how much is produced and removed)

GlodBlock:
i suggest you putting the clean code in `onRunningTick()` instead of `checkRecipe()`. otherwise it takes too much time to finish one clean cycle.

Dream-Master:
The filter vent casing is very cheap in Assembler

src/main/java/com/dreammaster/gthandler/multiAirFilter/GT_MetaTileEntity_AirFilterBase.java
boubou19: making an enum for 2 values seems a bit convoluted so i'm not sure about this one.

src/main/java/com/dreammaster/gthandler/multiAirFilter/GT_MetaTileEntity_AirFilterBase.java
boubou19: What do you mean? because it seems many multi do the same as i did.

repo-alt:
> 

`mMaxProgresstime` defines how often `checkRecipe` is called. It is definitely not needed to be invoked every tick

boubou19:
Well i can do so it runs every second instead of 10s, as the multi isn't overclockeable.

boubou19:
Ok so i added config entries, should give plenty of options to tweak their pollution cleaning rate. Also, now the multi cleans pollution every second instead of once every 10s. I tested it, and so far i have seen, it works well.

@Dream-Master indeed the multiblock cannot be used without turbines, so i let you decide if you want to move some small turbine recipes in assembler or if you want to gate the multiblock behind the MV assembler.

PolkageistHS:
Yeah, I was checking it too often, and the inflated numbers from the LV combustion engine were making me think it was the filter causing the pollution to go up.  It's working, although not knowing exactly which chunk it's currently working on for the current workload makes it a bit tougher to track down the actual numbers.  

Also, since the passive pollution spread between chunks happens while chunks are being cleaned, the poor scrubbers will be cleaning out a chunk, moving to the next one, and pollution will move back in to the first one, making the scrubber feel bad when it gets back to the first one only to see more pollution there again.  Not much to do about that though, unless you're cleaning your entire base instead of just the main polluted chunks.  

As far as the current recipes for the LV  go, it looks like the controller needs a minimum of 8 magnalium, so 3 magnesium + 6 aluminium.  Shouldn't be a problem, since you can get magnesium pre-TF from pulverizing/centrifuging marble cobblestone, and aluminium dust from alu gravel.  Everything else looks like it can be made in early/mid LV, other than the turbine, but that's outside this scope.

src/main/java/com/dreammaster/gthandler/multiAirFilter/GT_MetaTileEntity_AirFilterBase.java
bombcar: I think InteliJ is happier with enums when looking at the code from other places (the comments you have to go to to see)

bombcar:
> Well i can do so it runs every second instead of 10s, as the multi isn't overclockeable.

Is this one of the few that isn't? Will it refuse to build with a higher energy hatch because people might expect that it does.

src/main/java/com/dreammaster/gthandler/multiAirFilter/GT_MetaTileEntity_AirFilterBase.java
```diff
@@ -112,109 +141,151 @@ public boolean isFacingValid(byte aFacing) {
...
+                        }
                     }
                 }
             }
-        }
 
-        for (GT_MetaTileEntity_Hatch_Muffler tHatch : mMufflerHatches) {
-            if (isValidMetaTileEntity(tHatch)) {
-                mPollutionReductionWholeCycle += ((int) Math.pow(2.5, min(tTier, tHatch.mTier))) * 300;//reduction per muffler
+            ItemStack[] tInputs = Arrays.copyOfRange(tInputList.toArray(new ItemStack[0]), 0, 2);
```
Glease, line 208: Any reason to truncate this list?

src/main/java/com/dreammaster/gthandler/multiAirFilter/GT_MetaTileEntity_AirFilterBase.java
```diff
@@ -112,109 +141,151 @@ public boolean isFacingValid(byte aFacing) {
...
         mEfficiencyIncrease = 10000;
         mEfficiency = 10000 - (getIdealStatus() - getRepairStatus()) * 1000;
-        mPollutionReductionWholeCycle = 0;
-        mMaxProgresstime = 200;
-        mEUt=-32;
+        // check pollution for next cycle:
+        hasPollution = getTotalPollution() >= MainRegistry.CoreConfig.pollutionThresholdAirFilter;
+        mMaxProgresstime = getRecipe().mDuration;
+        mEUt=-getRecipe().mEUt;
         if(!hasPollution) {
```
Glease, line 174: Where goes the overclock handling? Also, isn't it supposed to consume less EU/t when it doesn't have pollution?

Dream-Master (approved):

src/main/java/com/dreammaster/gthandler/multiAirFilter/GT_MetaTileEntity_AirFilterBase.java
bombcar: It doesn’t overclock iirc

PR #239, opened 2021-12-07
use structurelib for air filter

GlodBlock:


boubou19:
Does it rotate the texture for the turbine as well?

GlodBlock:
it isn't allowed to rotate.

boubou19:
even better

Dream-Master (approved):

Dream-Master:
lets test this

PR #240, opened 2021-12-07
Distillation circuit

Dream-Master:
Rework the Pollution Filter recipes
Add new Distillation tower recipes
Needs GT PR 
* https://github.com/GTNewHorizons/GT5-Unofficial/pull/784


Glease (approved):
code looks ok. not sure about balance

PR #244, opened 2021-12-09
Asmd rework

Dream-Master:


Sphyix:
Looks fine for me

Glease:
Then tungsten smelting time needs tweaking. The old value is made based on the assumption that *you only need so many pure tungsten until all you need is tungstensteel*. Just adding lots of tungsten into the mix is a strict nerf.

Glease (commented):
Looks fine by me. (IV end, assline done) Not sure what those end game player think.

Sphyix:
Makes more sense to use tungstensteel,  I wasn't thinking at the smelt time, so good change.
Idk, I liked the recipes how they were initially on 2.1.1.0, but if a nerf is needed this looks fine.

GTNH-Colen (approved):

Dream-Master:
xsmd will follow later after the update

PR #247, opened 2021-12-13
add a oxygen

bombcar:
One of the proposals in [#9072](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/9072) is to nerf sio2 -> o2 only; removing it as a source of silicon.

This would NOT address the silicon ingot -> macerator -> silicon loop that @Sphyix mentioned, simply leave in the o2


Dream-Master (changes requested):
Where is the silicon? This recipe make zero sense.

bombcar:
> Where is the silicon? This recipe make zero sense.

It was removed so as to prevent early access to silicon

Dream-Master:
But this recipe make no sense at all

bombcar:
> But this recipe make no sense at all

ok will move it to the centrifuge and add back silicon

bombcar:
Fix https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/9072

Dream-Master:
The process time seems to be very short. Not sure about original recipe

bombcar:
old was LV 3 seconds at 30/t

new is MV 15 seconds at 60/t if I did the math right

Elisis:
Just leave the damn removal of SiO2 in and deal with it.

Dream-Master:
Let me check stuff ingame . I missing a few materials give si instead of si02

bombcar:
image

ten times the energy nice

Sphyix:
> ten times the energy nice

Why?
Making it take that much more energy is an INSANE nerf for low tiers. 
I cann totally understand doubling it, wtf is 10 times

Dream-Master:
Ok we can remove complete if more complains will be posted here

Sphyix:
> Ok we can remove complete if more complains will be posted here

It's your pack, you can do whatever you want, but people are complaining for a reason, some changes are wrong.
Go ahead and remove it, idc. I'll keep complaining if i feel like the change is dumb

Sphyix:
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/9072#issuecomment-991998820
You can read here for my reasoning for why it should be added back or slightly changed.
Post under there is correcting some of my calculations

Dream-Master (approved):

Savallator:
Also, why is this a centrifuge recipe? It does not make any sense... but since the increase in energy killed it for early-game Oxygen anyways, it does not even matter... guess people will the just use the electrolyzer with the different Oxygen recipes... for lategame silicon, the change does not even matter since you can use the MEBF for it anyways...

mitchej123:
Later game oxy & silicon aren't an issue, this was a nice QoL for oxygen early game, and some silicon for your first boules and such.  I'm not really sure what "problem" we were solving, sounds like a nerf in search of a problem 👎

Sphyix:
Please don't forget that this is the only way of getting silicon dust lategame as well 
The only other way would add an MBF and a macerator to get silicon dust at decent speed lategame

OrderedSet86:
Mostly I'm just confused about the purpose of the nerf. 

Is it a silicon nerf? Then only outputting oxygen is fine, no need to change EU.

Is it a realism change? If so, EU increase would make sense, but then it should be in the electrolyzer. Silica is not something that can be easily centrifuged. The way you would do this IRL is converting to a fluid first and then electrolyzing. If it is a realism nerf, then the EU should be increased and the electrolyzer recipe kept. https://www.quora.com/Is-there-a-chemical-reaction-you-can-use-to-separate-the-oxygen-out-of-sand-Silicon-Dioxide

Is the goal to remove the reliance on the electrolyzer for all dusts? If so, then keep the recipe unchanged in a centrifuge (although this would make no chemical sense).

I like the line for balance and aesthetic reasons, so I would like it to stay in, but I don't know what reasoning I should be targeting my suggestions at.

Savallator:
May i propose, to make silicon a bit more difficult, to have the need to purify it? Like, Wafers can no longer be made from silicon dust, but it needs to be purified to have almost no other atoms in it. For early game, it could be a centrifuge recipe with some loss, and later on a more complicated step involving the EBF. There could also be several "grades" of purity, where the normal wafers need only the low-grade (but yield more if made using better), and then the better wafers gradually need better silicon. It could be made in a way that it gives more yield, but with the tradeoff of needing more energy and complicated machines. In it's full extend, maybe we could get a silicon line like the NQ-line or platline. Maybe it could use HCl to make Trichlorosilane and then distill that in several steps like it is done in the real world, or in another step Silicon Difluoride (although Fluorine is way overused at the moment imho) like described [here](https://patents.google.com/patent/US4138509A/en)

Sphyix:
If you can think of a line that can still give me a stack/tick of silicon, sure..
If the speed ends up being like platinum in platline, this is a bad idea.
Silicon is needed in so large quantities that making it with a line like the naq or plat would kill a server on its own.

Alternative is to rework every recipe using silicon to use a lot less, like a TON less
I'd expect to use a stack of dust instead of a stack of blocks for t5 wafers

PR #249, opened 2021-12-15
remove small and tiny pipes out of the Planet dust

Dream-Master:
make recipe 36 times longer switch a tier up in eu/t and give a 10% discount on recipes
like 400 ticks x 30 eu/t x 36 -10%  = 3240 ticks at 120 eu/t...


bombcar:
None of the uptier breaks a requirement? they do look to still all be ... reasonable.

boubou19:
Stone dusts are totally optional, so it's indeed a buff

boubou19 (approved):

PR #251, opened 2021-12-18
Add bulk wood plate

bombcar:
Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/9172


Glease (commented):
You don't use a full amp. This means the machine cannot tolerate any wireloss, otherwise you need to provide a second amp which is a total overkill. I'd prefer 30 instead.

bombcar:
> You don't use a full amp. This means the machine cannot tolerate any wireloss, otherwise you need to provide a second amp which is a total overkill. I'd prefer 30 instead.

I always forget that - are there LV or something enums that would do that for us (be a bit below to allow for loss)?

Glease (approved):

PR #254, opened 2021-12-19
fixed  GTNewHorizons/GT-New-Horizons-Modpack#9106

boubou19:
fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/9106

Even if this PR fixes it, i find it a bit cumbersome to not rely on the translation manager to get the name of the blocks for the tooltip because it's not yet initialized or something.


draknyte1:
If you don’t want to hardcore it, your only option is to do as I suggested.

draknyte1 (approved):

boubou19:
i don't know when to update it, and i have too much stuff to do right now, so it won't be changed any time soon.

PR #262, opened 2022-01-09
add mixed metal ingot recipes in the assembler

boubou19:
added recipes of the mixed metal ingot into the assembler, so it reduces the need of molecular assemblers if you need to spam them:

image


bombcar:
> added recipes of the mixed metal ingot into the assembler, so it reduces the need of molecular assemblers if you need to spam them:
> 
> image

This shows three plates making 64 but that doesn’t seem to be what the code shows

boubou19:
i picked one of them to show, didn't want to post all the recipes.

bombcar:
Oh thats not copper obsidian and tin lol

draknyte1 (approved):

PR #274, opened 2022-01-31
port of AFSU, avaritia addons, iguana tweaks, computronics and architecturecraft minetweaker scripts

boubou19:


src/main/java/com/dreammaster/scripts/ScriptBiblioWoodsNatura.java
```diff
@@ -3,24 +3,26 @@
...
     public ScriptBiblioWoodsNatura() {
 
     }
 
+    @Override
+    public void initScriptData() {
+        scriptName.setLength(0);
+        scriptName.append("BiblioWoodsNatura");
+        dependencies.clear();
+        dependencies.addAll(Arrays.asList("BiomesOPlenty", "BiblioWoodsBoP", "BiblioCraft"));
```
bombcar, line 21: Doesn't depend on itself or Natura?

boubou19: good catch, bad c/p that i need to fix

bombcar (approved):

PR #286, opened 2022-02-25
Circuit0 fix

Dream-Master:
Change recipes using Circuit 0 in their recipes #9818

* https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/9818


src/main/java/com/dreammaster/bartworksHandler/BioItemLoader.java
```diff
@@ -116,7 +116,7 @@
                 new FluidStack[]{new FluidStack(BIOFLUIDS[3],50)},
                 1200,BW_Util.getMachineVoltageFromTier(8),100,8,0,false
         );
-        for (int i = 0; i < OreDictionary.getOres("cropTcetiESeaweed").size(); i++) {
+        for (int i = 1; i < OreDictionary.getOres("cropTcetiESeaweed").size(); i++) {
```
Elisis: Shouldn't this be .size() + 1? Otherwise you miss out on looping with the final ore

PR #296, opened 2022-03-11
Neutron Reflector Graphite Sanity Reduction

ScriptedPiky:
Halves the amount of Graphite that is required, as even if you took the cheapest route, the Tungstencarbide recipe, will take a **minimum** of 256 Graphite for a single Iridium Neutron Reflector. Or in this case, if you wanted to make a fusion T2 or T3 reactor (with the one included for the reactor itself), it'll take 16896 Graphite.


LewisSaber:
There is enough graphite with Void miner

ScriptedPiky:
I'm not going to balance around a Void Miner, it's rather silly to do so

LewisSaber:
VM is already a part of game thing that everyone uses to get resources, it means that game should be balanced around them

Glease:
VM is an ungraceful hack to fix endgame. Anything but endgame should just ignore it outright when balancing.

boubou19:
> VM is already a part of game thing that everyone uses to get resources, it means that game should be balanced around them

well then if it's basically using endgame stuff to justify any nerf which is silly.

VM right now is unbalanced as fuck because you can make Og trivially. A rework in the creation of all the boosting gases is needed. Oh and we should definitely yeet replicator recipes for the gases.

Dream-Master:
I think Graphit is not a issue. You get tones of it

ScriptedPiky:
Not in the range that would be these current values though. I tried making two T2 fusion reactors and already used up like 32k, if not more. A good portion is already used in IV, I'm not sure about the Graphene/Graphite in the MT situation, and mining them in diamond veins isn't that great of a resource. Although, there is the GT++ method of mass production, but that is perhaps in its own tale & questions of balance. There's also the Blood Magic meteors, but that also requires player to actually go through a good portion of BM itself, so that is off the table as well.

Dream-Master (approved):

PR #318, opened 2022-05-30
buff multifiberglass and bio circuit board

GlodBlock:
- elite circuit board is using 8 platium foils instead of 16
- multifiberglass circuit board is using palladium foil instead of platium
- bio circuit board is using 24 nt foils instead of 48

the crystal circuit is using more platium than quantum one which make ppl stick to quantum instead of upgrade their line.
the bio circuit is too expensive and nobody will use it unless you must need uev mainframe. now cutting down usage of the nt foils


GTNH-Afx237v7:
looks good to me, especially finally some use for palladium.

miozune (approved):

PR #330, opened 2022-06-26
DTPF Recipes

Connor-Colenso:


src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
```diff
@@ -2063,189 +2063,212 @@ public void run2() {
...
+                        new ItemStack[]{GT_Values.NI},
+                        new FluidStack[]{Materials.DimensionallyTranscendentResidue.getFluid(fuel_quantity_2), Materials.Infinity.getMolten(128L * 144L)},
+                        base_time/2, 409_600_000, eternal_heat);
+            }
 
-        if (Loader.isModLoaded("miscutils")) {
+            // UV Superconductor.
+
+            {
+                long base_quantity = 2043L;
```
Pnc6lx, line 2261: why 2034? it is not a multiple of 144, is it 2304 (16*144)?

Pnc6lx, line 2261: so why is it 2043? it is a strange number because it is 9*227, maybe it was 2304?

Dream-Master (approved):

src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
Connor-Colenso: Because 2043 is the closest number to 2048, below 2048 which is also a multiple of 9 (rounded down basically).

This is because there are 9 ingredients going into the recipe, otherwise there would be non-144 multiples of molten fluids going in. Other than the actual fuel material, which is precisely calculated to give a certain EU discount for each recipe.

PR #341, opened 2022-07-18
Silicon smelting fixes

chochem:
Added chem bath cooling for silicon and siliconSG.

revert to using siliconSG plates for transistors again.

image

addresses https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10748 and https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10686 

please dont merge without dream seeing it. :)


ScriptedPiky:
How come you end up lose 3/4ths of the coolant? And is this the primary way of obtaining it or is there another way? As this seems rather expensive in time, EU/t, and time to make coolant (this stuff takes forever to make)

chochem:
yeah I just copied the kanthal recipe and reduced the time/eu a bit. :)
The time and eu cost is not that much compared to the EBF recipe. But feel free to make some balancing suggestions if this is overall too much for MV.

This is the primary way until you reach a vaccum freezer. Only alternative would be SMD Transistors (which need MV assembler and gallium).

PR #344, opened 2022-07-20
Add EBF Recipes to Space Ores for Turbines

Steelux8:
- Added EBF recipes for the ores that were buffed in the respective PR in GT5-Unofficial.

Meant for [#1146](https://github.com/GTNewHorizons/GT5-Unofficial/pull/1146). The new ores which had their turbine numbers greatly buffed will get EBF recipes, with the same time per ingot as Oriharukon, but an appropriate EU/t and coil tiering for each one.


Dream-Master:
the process time is very short compared to the heat values. Is this balanced compared to other materials ? @Steelux8

Steelux8:
These are optional materials, for turbines only. Do they have to match processing times of required materials for the tier?

Dream-Master:
Idk I just asking.

Steelux8:
I kept the processing time from Oriharukon, but with higher EU/t and coils. If other values are better, I'm willing to change them.

Dream-Master:
ok lets try this and if we think it is unbalanced we change it later

Dream-Master (approved):

PR #347, opened 2022-07-22
Standardize all electric components to use 1s LV assembler

OrderedSet86:
1. Motors, pumps, etc. all follow different conventions on how long their assembler time should be. This standardizes them to 1s LV. This is fine because of limited usefulness of the recipes and because motors were already LV at every tier and pumps were already 1s at every tier.
2. The assembler recipes are mostly pointless - either you make them by hand in a worktable or your AE system makes them in a molecular assembler. The only mod this should affect is Project Red: Transportation and possibly Logistics Pipes.


OrderedSet86:
Solves https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10751

Connor-Colenso (approved):
Seems reasonable.

Dream-Master:
are you have any benefit to craft it via assembler (material wise)? If not all good if you have better ratios then we should tier it.

OrderedSet86:
@Dream-Master No efficiency benefits, they're the same as the hand crafting recipes.

PR #350, opened 2022-07-27
Make Aluminium Hydroxide actually craftable at MV

miozune:
https://chemistry.stackexchange.com/questions/25045/how-to-convert-sodium-aluminate-back-to-sodium-hydroxide

Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10774


Dream-Master:
why we not add a circuit. If we add a second recipe with this two input items/fluids we run in issues @miozune

Dream-Master (approved):

PR #367, opened 2022-08-09
add craft to infinity egg from witchery

POPlol333:
as I've been asked here https://github.com/GTNewHorizons/Avaritia/pull/21


chochem:
what is the plan for the quest then? Also, why not use some of its ingredients?

for reference: 
image

chochem:
talked on discord. all good, go ahead.
was confused by the 'add'. this is just a change. :)

PR #371, opened 2022-08-15
Moves AE2FC recipes to coremod and made them more greggy

Runakai1:
Discretizer:
image
Pattern Encoder:
image
Pattern Decoder:
image
Dual Interface:
image
Ingredient Buffer:
image
Large Ingredient Buffer:
image
Fluid Import:
image
Fluid export:
image
Fluid P Terminal:
image
Processing terminal:
image

Cell recipes in CA are the same as the normal item ones, just with more pumps.
image

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -8,21 +8,8 @@ dependencies {
     compile("com.github.GTNewHorizons:Avaritia:1.23:dev")
     compile("net.industrial-craft:industrialcraft-2:2.2.828-experimental:dev")
     compile("com.github.GTNewHorizons:GTplusplus:1.7.62-pre:dev")
-    compile("com.github.GTNewHorizons:AE2FluidCraft-Rework:1.0.6-gtnh:dev")
 
     compileOnly("com.github.GTNewHorizons:Hardcore-Ender-Expansion:1.9.1-GTNH:dev") {
-        transitive = false
-    }
-    compileOnly("com.github.GTNewHorizons:AE2FluidCraft-Rework:1.0.6-gtnh:dev") {
-        transitive = false
-    }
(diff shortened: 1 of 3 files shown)


miozune (commented):
I think we should not add more and more mods in dev env unless necessary, it slows down loading time. (Or is it asked by someone?)
- `Config.fluidCells` and `Config.fluidIOBus` are `true` by default
- `ItemBasicFluidStorageCell` and `CellType` can be replaced by `getModItem`
- `ModAndClassUtil.OC` can be replaced by `Loader.isModLoaded("OpenComputers")`

Also use `getModItem` instead of `GameRegistry.findItem` or `GameRegistry.findItemStack`. The latter 2 also work but be consistent with other scripts

dependencies.gradle
```diff
@@ -8,10 +8,23 @@ dependencies {
     compile("com.github.GTNewHorizons:Avaritia:1.23:dev")
     compile("net.industrial-craft:industrialcraft-2:2.2.828-experimental:dev")
     compile("com.github.GTNewHorizons:GTplusplus:1.7.62-pre:dev")
+    compile("com.github.GTNewHorizons:AE2FluidCraft-Rework:1.0.6-gtnh:dev")
 
     compileOnly("com.github.GTNewHorizons:Hardcore-Ender-Expansion:1.9.1-GTNH:dev") {
         transitive = false
     }
+    compileOnly("com.github.GTNewHorizons:AE2FluidCraft-Rework:1.0.6-gtnh:dev") {
```
boubou19, line 16: why adding it two times? also if it's just added so you can see the recipe in dev, that is not really required: the script if done correctly will load itself only if all the mods involved in the recipe are present. So you can put an AE2FC dev jar in your dev env's mods folder instead.

boubou19 (changes requested):
Please use only getModItem to be consistent with the other scripts.

boubou19 (approved):

PR #372, opened 2022-08-16
Disassembler Dupe Blacklist

Connor-Colenso:
More exploits from this very useful machine...

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,7 +1,7 @@
 // Add your dependencies here
 
 dependencies {
-    compile("com.github.GTNewHorizons:GT5-Unofficial:5.09.40.87:dev")
+    compile("com.github.GTNewHorizons:GT5-Unofficial:5.09.40.96-pre:dev")
     compile("com.github.GTNewHorizons:Yamcl:0.5.82:dev")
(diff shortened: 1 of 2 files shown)


src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
```diff
@@ -1719,6 +1752,15 @@ private void makePlasmaForgeRecipes_DTPF() {
                         new FluidStack[]{Materials.DimensionallyTranscendentResidue.getFluid(1000)},
                         40*20, 128_000_000, eternal_heat);
 
+
```
Dream-Master, line 1755: why you ship this in the dissasembler pr ?

dependencies.gradle
```diff
@@ -1,6 +1,7 @@
 // Add your dependencies here
 
 dependencies {
+    compile("com.github.GTNewHorizons:GT5-Unofficial:5.09.40.96-pre:dev")
```
mitchej123, line 4: Two versions?

mitchej123: Also no pre's!

Dream-Master:
?

PR #375, opened 2022-08-17
Add assembler recipes for Advanced Storage Housing and Fluid Storage Housing

dipo33:
This PR adds assembler recipes for **Advanced Storage Housing** and **Fluid Storage Housing** to remove the necessity of using tools.

Same assembler recipes already exists for the **ME Storage Housing**.

Closes issue https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10265

Also fixes previously wrongly named variable
image
image

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -66,7 +66,5 @@ dependencies {
     compileOnly("com.github.GTNewHorizons:BuildCraft:7.1.26:dev") {
         transitive = false
     }
-    compileOnly("com.github.GTNewHorizons:ExtraCells2:2.5.14:dev") {
-        transitive = false
-    }
+
 }
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -66,5 +66,7 @@ dependencies {
     compileOnly("com.github.GTNewHorizons:BuildCraft:7.1.26:dev") {
         transitive = false
     }
-
+    compileOnly("com.github.GTNewHorizons:ExtraCells2:2.5.14:dev") {
+        transitive = false
+    }
 }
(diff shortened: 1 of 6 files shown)


Runakai1 (changes requested):
Move the recipes to AE2FC script. We don't need more dependencies in coremod-
No need for conditions to see if mod is loaded then. 
Also, why the edited file with spaces?

dipo33:
- I was told we are trying to move to code recipes instead of scripts, are you sure we want new script recipes? Also, didn't you mean ExtraCells mod instead of AE2FC ? The housings are from the ExtraCells, not AE2FC.  I would agree that moving it to ExtraCells would probably be better, if you did mean that.
- The edited spaces are my fault, IDEA automatically deletes all trailing space on file save, I will probably have to disable it. 
- Will add the configured circuit.

Runakai1:
> * I was told we are trying to move to code recipes instead of scripts, are you sure we want new script recipes? Also, didn't you mean ExtraCells mod instead of AE2FC ? The housings are from the ExtraCells, not AE2FC.  I would agree that moving it to ExtraCells would probably be better, if you did mean that.
> * The edited spaces are my fault, IDEA automatically deletes all trailing space on file save, I will probably have to disable it.
> * Will add the configured circuit.

Yes we're moving from scripts to code. Inside coremod there are classes named ScriptAlveary e.g. You can make another class or simply add to AE2FC, I just recommended it, since it already builds upon extra cells dependency.

Runakai1 (approved):
Now that it is in its own file and it looks properly done I think its good to merge. Sorry for all the oscillation.

PR #376, opened 2022-08-18
gt++ should probably be a compileOnly thing

Glease:


Dream-Master:
Why the build test fails ?

Glease:
because colen forget to wrap stuff related to gt++ in `if(gtpp_loaded)`

Glease:
Yeah and now it works.

Dream-Master (approved):

PR #393, opened 2022-08-28
Rework Wooden Form

Quarri6343:
This PR overhauls Wooden Form:

・Make Wooden Form recipe shapeless.
It made many new players complain they can't craft it
We have to remove https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/blob/0aadf25ed1399a89a8f199c801f279e53912a46b/scripts/CoreMod.zs#L864 after this

・Wooden Form now has 100 durability.
Since Wooden Form is a crafting tool, it should be slowly consumed on craft like other GT tools.
The original idea is from https://github.com/GT-IMPACT/Impact-Core/blob/b8559ea8dc9fafccb30e6b852717a6e110476eee/src/main/java/com/impact/common/item/WoodBrickFormTool.java

・Remove broken shapeless Unfired Coke Oven Brick recipe
It is not working in current release so it is safe to remove i guess

https://user-images.githubusercontent.com/39122497/187077289-6a6ef972-e7cb-4502-80fe-64e3f92d8e70.mp4


Dream-Master:
cool idea. Is it possible give the form a bar to see visual how much duribility left ? @Quarri6343

Quarri6343:
I'll give it a try.

Quarri6343:
it should look better
https://user-images.githubusercontent.com/39122497/187081335-e41defc2-ed10-44d1-8ba4-d85eda9fb7a6.mp4

Dream-Master:
cool

src/main/java/com/dreammaster/item/WoodenBrickForm.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private int getNBTDurability(ItemStack pItemStack) {
+        createOrInitNBTTag(pItemStack);
+        return pItemStack.stackTagCompound.getInteger(NBTTAG_DURABILITY);
+    }
+
+    @Override
+    public ItemStack getContainerItem(ItemStack stack) {
+        ItemStack ret = stack.copy().splitStack(1);
+        int tCurrentDura = getNBTDurability(ret);
```
miozune, line 98: Why you need NBT when you can get durability from item damage?

Quarri6343: Because Oven Glove implemented it in this way
If there is no reason to use it I can delete it
https://github.com/GTNewHorizons/NewHorizonsCoreMod/blob/69a50027401c1ed34e93ffee4b010f2fa69d0773/src/main/java/com/dreammaster/baubles/OvenGlove.java#L171-L173

miozune: Yes please

miozune: Gloves use item damage to distinguish left / right hand

miozune (approved):

PR #399, opened 2022-08-30
add a Diode recipe using Wafer and Tantalum wire

Dream-Master:


Dream-Master:
This is for the Diode not the SMD

Glease (approved):
Looks ok.

PR #401, opened 2022-08-30
removed unused item recipe from Netherstarfragment

Dream-Master:
This Item has no use any longer.


Dream-Master:
Not used in any recipe so i merge this now

PR #408, opened 2022-09-03
Fix Glass Fiber Wiremill Recipe

LewisSaber:
There is no reason to make player have entire extruder(or bus) for 1 recipe


LewisSaber:
Should have added Circuit №1

Dream-Master:
no fine wires use 3 or is this a normal wire ? if so use circuit nr 1

Dream-Master (approved):

PR #413, opened 2022-09-12
Change Wooden Brickform durability

Dream-Master:
Change Wooden Brickform durability to 10 stacks (640)

### DIFF
--- a/src/main/java/com/dreammaster/item/WoodenBrickForm.java
+++ b/src/main/java/com/dreammaster/item/WoodenBrickForm.java
@@ -17,7 +17,7 @@ public class WoodenBrickForm extends Item implements IExtendedModItem<WoodenBric
 
     private final String _mCreativeTab;
     private final String _mItemName;
-    private static final int maxDurability = 100;
+    private static final int maxDurability = 640;
 
     @Override
     public WoodenBrickForm getConstructedItem() {


Quarri6343 (commented):
I think it is too high
You can craft 3 x 640 / 4 = 480 unfired coke oven brick"s" with 640 durability brick form
According to the quest book which requires 600 coke oven bricks, including seared brick usage, it means you only need two brick forms to beat the pack
If you want this high durability, you may want to revert the entire durability thing...

Dream-Master:
You have to craft more than one coke oven and normal bricks as well for steam machines. @Quarri6343 suggest a better durability.

dragagon:
There is also the 24 coke oven quest for benzene, but you do have access to the alloy smelter at that point. I see one of two things. either keep it at 100 and remove the single brick recipe or keep it at something like 640 and remove the 8x wrap recipe. As it stands now, 100 is too low if the player doesn't realize there is a durability and does the easy shapeless craft, but I can see how 640 is too much if they have access to the 8x shaped crafting.
image
image

Quarri6343:
I agree
I think questbook should encourage people to do efficient 8x wrap recipe instead of removing shapeless recipe

you can craft 480 / 26 = 18.5 coke ovens from 640 durability brick form...

dragagon:
I can agree with that, was a shock when it changed and the quest book didn't mention it has a durability. and being crafted from a tinkers pattern would make a player assume it doesn't have a durability as none of those forms have them either. 

I would suggest a compromise, and say something like... 300 crafts and a note added to the Tier 0 - Stone Age, Forming Press v0.1 warning that it has a durability.

Dream-Master:
@dragagon its 300 now

Connor-Colenso:
> If you want this high durability, you may want to revert the entire durability thing...

Sounds good to me.

PR #425, opened 2022-10-08
Indusctors

Dream-Master:
Add Inducotor recipes

* https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/11579


src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
```diff
@@ -16973,6 +16973,27 @@ public void run3() {
                 ItemList.Circuit_Parts_CapacitorSMD.get(48L),
                 320,
                 120);
+
+        GT_Values.RA.addAssemblerRecipe(
+                new ItemStack[] {
+                    GT_OreDictUnificator.get(OrePrefixes.ring, Materials.Neodymium, 1),
+                    GT_OreDictUnificator.get(OrePrefixes.wireFine, Materials.Platinum, 2),
```
GlodBlock, line 16980: platinum is already used for diode, also it is also heavily used in other circuit compents. personally i suggest to use fine conductive iron wire

GlodBlock:
also this issue is related https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/11579
@NotAPenguin0 how do you think of replacing platium with something else

src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
```diff
@@ -17075,6 +17108,18 @@ public void run3() {
                 ItemList.Circuit_Parts_DiodeXSMD.get(64L),
                 160,
                 120000);
+        // Inductor
+        GT_Values.RA.addAssemblerRecipe(
+                new ItemStack[] {
+                    GT_ModHandler.getModItem("miscutils", "itemFoilTitansteel", 4L),
+                    GT_ModHandler.getModItem("miscutils", "itemFoilBotmium", 1L),
```
GlodBlock, line 17115: botmium and titansteel are the uv tier alloy while all other xsmd material is zpm/luv tier. imo it is gated too later. i suggested to use artherium-sn and hikarium foil which is a zpm/luv material

src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
NotAPenguin0: Replacing platinum could work, the only concern is that the recipe should be in line with the cost of the other SMDs. Maybe if it gets too easy using e.g. fine conductive iron the yield should be 24 and 48 instead?

src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
NotAPenguin0: Consistency with the other xsmd components is the most important thing. I wasn't sure what to use as I've never played with any of these materials/components. If you think those are better, I'm all for it.

src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
GlodBlock: smd is unlocked in mv so i don't think we should make it too difficult. but yea conductive iron outputing 32 inductor is too much. conductive iron with 24 inductor should be more suitable

NotAPenguin0:
When the recipes are finalized we should also add recipes to craft circuits using these new SMD components (unless those are automatically generated, I'm not 100% familiar with the codebase).

GlodBlock (approved):

GlodBlock:
> When the recipes are finalized we should also add recipes to craft circuits using these new SMD components (unless those are automatically generated, I'm not 100% familiar with the codebase).

the usage is added in https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/424

PR #432, opened 2022-10-11
Prevent Dynamism Tablet from getting custom mob drops

kuba6000:


kuba6000:
Well these were always supposed to be player only. 
But for some reason Thaumcraft decided to make its own fakeplayer instead of using the existing one

boubou19 (approved):

Steelux8:
If I had seen this, I would have contested it right away.

First off, why? We're still approving PRs with no explanation at all, this is not how it should be,

Second off, even though the farms that can be done with this are good, I didn't notice any outputs that are really overpowered or that allow tier-skipping. On top of that, this is magic-related, so I'd wager it's good for the pack if this is a good automation option. Is it really the intention that these drops have to farmed manually, in a pack all about automation?

Viktor-KN:
I have to agree with @Steelux8 Why this custom mob drops even exist then, if don't have a way to automate this? In pack about automation no one will be spending hours to get this loot by hand... Please return this magical way to get it automated, or make new one (with EEC perhaps?).

RaitGamerGR:
Additation to @Viktor-KN's suggestion, what about mode for EEC with special drop? Mb with more energy consumption or liquid xp consumption?

theMegaTech:
Agreed about EEC, what's the purpose of 'lagless mobfarm' if u still need to use regular spawners to get "player-only" drops? For now EEC is only useful for BM and regular mobdrops, but if u wanna farm anything more useful than, for example, enderpearls - u need to setup mobfarm that still lags. Why don't either remove player-only drops at all or add them to EEC?

chochem:
player-only drops are supposed to be bonus things you get for actually fighting yourself. They are not supposed to be farmable.

theMegaTech:
but who would "fight" (funny word in pack with TCon, imagine "fighting" after getting crossbow) in pack about automation? and they would be farmable anyways, i don't think there's no other fakeplayer option. Why not remove that special loot at all then? Or maybe let anyone farm it cuz anyways what's so bad in having another option to get infinite loot from nothing. It anyways isn't more powerful than other effortless options like crops or bees, but nobody saying that crops/bees are bad

Viktor-KN:
> player-only drops are supposed to be bonus things you get for actually fighting yourself. They are not supposed to be farmable. (at least thats the idea here, there might be some options left)

Then reward should be much much more better then we have now. If its not automateable... well, most of this drop may be considered as junk, occupying inventory slots, especially tiny piles of whatever it drops...

boubou19:
if you check the code, it is clear that fakeplayers are meant to be excluded. The fact that the dynamism tablets have a custom fake player was an oversight and is now corrected.

The drops are meant to be nice random loots for early tf explorers. Relying on those loots is not intended, but if you want to grind for them, up to you. In any case, it is really not meant to be automatable.

theMegaTech:
> The drops are meant to be nice random loots for early tf explorers

but sadly it isn't cuz just by exploring you get to kill like 3-4 redcaps for example and their loot is awful in small amounts (but amazing if u killing thousands of 'em (this is farming anyways))

Viktor-KN:
@Dream-Master we need your opinion on this change...

Steelux8:
After talking with the other devs, I do believe that Dream is the one who should chime on this. I heard that the dynamism tablet has other problems I didn't know about, but the whole idea of "items that only drop from manual player kills" is what should be discussed here. If he says that these items could be automated, then both the dynamism tablet and the EEC should drop them.

mitchej123:
It was not what was intended by dream. 

Also what was originally intended by mod authors not part of gtnh, is irrelevant. The only thing that matters is what we and dream decide. 


Of all the hills to die on, this seems an expecially odd one.

cubic-chariot:
> "Many players cannot help approaching a game as an optimization puzzle. Given the opportunity, players will optimize the fun out of a game." - Soren Johnson, Civilization series designer

I personally never liked the idea of player-only drops, precisely because it is hard to automate, but it is not challenging; the difficulty lies in finding these exploits rather than designing the farm.

If the goal is to give a boost to early TF explorers, one option would be to add e.g. gallium to loot chests (which is by design not repeatable).

Dream-Master:
@kuba6000 but in the EEC this drops are possible ?

kuba6000:
> @kuba6000 but in the EEC this drops are possible ?

No, they aren't

PR #434, opened 2022-10-16
Replace Poison Ivy with cacti in upgrade

Runakai1:


kuba6000 (approved):

boubou19:
this is a really good change. I spent like a whole week travelling on delta's map just to find like one place where there are some.

PR #439, opened 2022-10-26
Buff Tantalum

Dream-Master:
Reduce EBF and cooling time for Tantalum


Connor-Colenso:
I honestly feel at a minimum this should become an LV EBF recipe with no cooling required. I just don't get the logic of making it so high, it renders EIO even more obsolete.

Dream-Master:
> I honestly feel at a minimum this should become an LV EBF recipe with no cooling required. I just don't get the logic of making it so high, it renders EIO even more obsolete.

No. It need 3000 °C to smelt. The smd parts use it and if we change that i will change all this parts to a higher tier material so tantalum will becomes obsolete.

Dream-Master:
I will add a Capacitor pr in a few mins change Ender io Capacitors use alu blots for basic version

Dream-Master:
This change is not needed. I oredict the silver capacitors and the problem is solved

PR #443, opened 2022-11-06
remove pure alu to sodium alunite recipes

Dream-Master:
change sodium alunite recipe gives always 64 items per craft change red mud amount per craft
change red mud recipe from 10k to 1K
add the sodium alumitae mixer recipes to  chem reactor recipes (pure variants)

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
+++ b/src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
@@ -5049,9 +5049,9 @@ public class GT_MachineRecipeLoader implements Runnable {
                 480);
         GT_Values.RA.addMultiblockChemicalRecipe(
                 new ItemStack[] {
-                        GT_OreDictUnificator.get(OrePrefixes.dust, Materials.Mica, 16L),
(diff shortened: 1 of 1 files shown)


src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
```diff
@@ -5055,17 +5023,17 @@ public void run() {
...
                     GT_Utility.getIntegratedCircuit(24)
                 },
-                new FluidStack[] {Materials.Water.getFluid(1000L)},
+                new FluidStack[] {Materials.Water.getFluid(2000L)},
                 new FluidStack[] {
-                    Materials.Hydrogen.getGas(100L),
-                    Materials.CarbonDioxide.getGas(1750L),
-                    Materials.RedMud.getFluid(500L)
+                    Materials.Hydrogen.getGas(1000L),
+                    Materials.CarbonDioxide.getGas(2000L),
```
BlueWeabo, line 5033: The hydrogen seems fine.

It should be more equal towards the doubling, which would mean you should make it 3000L or 4000L.

Dream-Master: Done

BlueWeabo (approved):

PR #453, opened 2022-12-12
Update JABBA Iron Upgrade to remove circuit conflict 

BritishCynic:
Resolve circuit conflict by changing the circuit used from 1 to 6 so it does not conflict with other recipes involving the components. Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/11326


Ethryan (changes requested):
Hmm, maybe transfer all upgrade recipes to a different circuit to let them keep consistency.
Just editing the conflicting one makes it stand out and behave strangely if you are used to them all having the same circuit configuration.

BritishCynic:
Circuit 13 looks viable if you'd like all JABBA upgrades moved to that circuit

BritishCynic:
Gonna close this one and then do the circuit 13 changes.

PR #455, opened 2022-12-16
Make MV cooling lossless on coolant

chochem:
Problems:
1. MV cooling is a pretty bad mandatory time sink. I think the bit of complexity is very nice (ic2 coolant, potentially cooling the hot coolant back down). And some difficulty here makes the progression to the vac freezer feel good. But currently the time sink is over the top imo. It literally takes 2 hours of cooling (with one machine) to upgrade one EBF to kanthal. Then ofc there is Silicon and potentially Tantalum cooling. Kanthal cooling being annoying and over the top is the most common feedback from all players that have recently played MV.

2. The 75% loss of coolant is pretty extreme compared to other lossy processes. If you lose so much there is no incentive to try to invest in the recovery process.

3. its unclear why the cooling times for the 4 ingots should be so different. (but this is not that important)

Solution:
This PR reduces and unifies all the cooling times to 20 seconds and the loss to just 25%. This makes the times more appropriate to the tier and the 'hot coolant' cooling more interesting.

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
+++ b/src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java
@@ -7549,7 +7549,7 @@ public class GT_MachineRecipeLoader implements Runnable {
                 GT_Values.NI,
                 GT_Values.NI,
                 null,
-                600,
+                800,
                 120);
 
         // Mysterious crystal upgrading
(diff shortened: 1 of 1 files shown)


chochem:
ok. After some discussion with dream and steelux I removed the time buffs. However they agree with the coolant loss and both suggested I make it 250 in and out. THat is now all the PR does.

Ethryan (approved):

PR #459, opened 2022-12-21
Fix melodic capacitor recipe

chochem:
Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/12057

tested in full game and works.


miozune (approved):
`new ItemStack(Items.nether_star, 1, 0)` should work, but either is fine I guess

PR #473, opened 2023-01-01
change 300s to 180s

POPlol333:
Xenoxene as now is 300s at UEV, slowing the whole radox line to ground and it's honestly annoying to send that much energy into a culture that is already really dumb rng based to get slowed by it


Phineasor (approved):
Looks good will be super helpful, this is the largest bottleneck on Radox, and Radox is needed more mid game now for boards.

PR #482, opened 2023-01-09
Move all TC researches, recipes from scripts

kuba6000:
CODE IS DONE (I THINK, the game starts fine)

- [x] TRANSLATIONS

https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/13232


Dream-Master:
works for me

PR #483, opened 2023-01-15
Add recipes to the nei planet blocks

GDCloudstrike:
Adds recipes for the nei planet blocks as they are used in EOH recipe selection
They are tiered to fit their corresponding rocket tier and can be crafted after visiting their respective planets

Example:
image


Dream-Master (changes requested):
please add control circuits

Dream-Master (approved):

PR #484, opened 2023-01-16
Fix compressed chest conflict

chochem:
Fix one of the recipe conflicts in https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/12300 between compressed chest and obsidian chest.

also reduced the dense obsidian to be in line with the other recipes. Prompted by
image

new recipe:
image


Dream-Master:
I am not sure if this is Material wise balanced need to check ingame. (Compared to shaped recipes)

Dream-Master (approved):

PR #487, opened 2023-01-20
Fix catalyst amount for chromatic glass dtpf recipes

GDCloudstrike:
I did a minor calculation mistake in the last pr, this fixes that

old values
image

new values
image


BlueWeabo:
I was wondering why at first it was the same eut price in both methods to make it but produced more.

BlueWeabo (approved):

PR #493, opened 2023-01-25
add covert recipe between ec and ae large cell

GlodBlock:
a shapeless recipe to turn ec item cell in ae2 one. the cell's storage contents can be kept so no need for using me-io to transfer. but the filter and upgrade will be lost


LewisSaber:
What if you also add tooltip on ec2 cells that content will be kept?

minecraft7771 (approved):

PR #501, opened 2023-01-31
Sanitize recipes

boubou19:
same as https://github.com/GTNewHorizons/GT5-Unofficial/pull/1688 but for NHCore :D


BlueWeabo (approved):
Only one way to find out if something broke

PR #504, opened 2023-02-02
Update NameRemover.java

BritishCynic:
This adds new functionality to the Name Remover to remove additional NBT tags. Tested in-game and functionality is exactly as expected.

* Can now remove dyes from armor by extending the existing removal of the "display" NBT tag to also cover the removal of "color" tag at the same time if present.
* Can now remove the "RepairCost" NBT tag that is placed on items repaired by the Anvil (Repaired armor will not stack with un-repaired armor pieces in AE2)
* Can now remove the "color" NBT tag used when Railcraft tank blocks are removed from the world (causes blocks to not stack with those yet to be placed in both regular inventory and AE2)

Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/12384 and https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/9986


src/main/java/com/dreammaster/gthandler/nameRemover/NameRemover.java
```diff
@@ -81,20 +81,35 @@ public int checkRecipe() {
...
                 default:
                     removeName = true;
                     removeDisassembly = true;
+                    removeColor = true;
+                    removeRepair = true;
             }
 
             if (removeName && nbt.hasKey("display")) {
                 nbt.getCompoundTag("display").removeTag("Name");
+                nbt.getCompoundTag("display").removeTag("color");
```
mitchej123, line 99: hmm, this won't properly remove display.color if you pass in `removeColor`.  Should change the top level check to `hasKey(display)`, and then each individual check for removeName, and removeColor

BritishCynic:
So, to address Mitch's comments in a discussion yesterday:

Yes, this does make repairing vanilla weapons and armor easier, _however_ I don't think this really changes much in the grand scheme of things:

The cost set by the anvil goes up in terms of player levels, not raw materials. TiCon doesn't penalise its players in XP to repair its tools, and while this change would still result in players still needing to expend XP for repairs, it makes repairs for armor comparable to what TiCon offers for tools.

src/main/java/com/dreammaster/gthandler/nameRemover/NameRemover.java
```diff
@@ -120,7 +151,11 @@ public String[] getDescription() {
         description.add(UNDERLINE + "Second Slot" + RESET);
         description.add("One of the following circuits:");
         description.add(BOLD + "Circuit 1:" + RESET + "  Attempt to fix broken ores by removing the Display Name tag");
-        description.add(BOLD + "Circuit 2:" + RESET + "  Remove disassembly tags");
+        description.add(BOLD + "Circuit 2:" + RESET + "  Remove GT Disassembler tags");
+        description.add(BOLD + "Circuit 3:" + RESET + "  Remove Railcraft stacking tag");
```
mitchej123, line 155: Any reason not to just do circuit 3, 5, and 6 as one "remove color"?

mitchej123 (approved):
Seems fine as is.  I'd suggest making one circuit to do color, dye, and spray instead of 3 separate circuits, but no strong opinon.

PR #518, opened 2023-02-11
Move some GC script to coremod + add rocket parts assembler recipes

LewisSaber:
Moves Crafting Rable recipe removal to coremod
Add Assembler Recipes For Rocket Parts
Add Crafting recipes of some gc parts that are defined in script
Script tweaking PR will follow soon


greesyB:
IMO GC recipes should either stay in GC as it has the necessary dependencies to register them already, or at the very least never be registered in GC to avoid registering, removing, registering again. I started this work a while back but haven't yet had the time/motivation to finish. That branch is here - https://github.com/GTNewHorizons/Galacticraft/pull/72

LewisSaber:
ig this pr is useless now?

Dream-Master:
@greesyB we put it in core mod for now.

PR #524, opened 2023-02-15
cleaned up alloy smelter recipes

boubou19:
This PR aims to make the recipes look clean. This is a test PR, by that i mean that once we all agree to the style, i'll (very slowly) migrate all the recipes to this style.

So please make all the remarks you can, because i'm not going to migrate to a different style multiple times 😂


POPlol333:
Isn't putting literally everything in variable like you are proposing to kind of overkill ?

src/main/java/com/dreammaster/gthandler/recipes/AlloySmelterRecipes.java
```diff
@@ -8,236 +11,343 @@
...
+
+        ItemStack x1_ObsidianBlock = new ItemStack(Blocks.obsidian, 1, 0);
+
+        ItemStack x1_ObsidianIngot = GT_OreDictUnificator.get(OrePrefixes.ingot, Materials.Obsidian, 2L);
+
+        ItemStack shapeMoldIngot = ItemList.Shape_Mold_Ingot.get(0L);
+        ItemStack shapeMoldBall = ItemList.Shape_Mold_Ball.get(0);
+
+        GT_Values.RA.addAlloySmelterRecipe(
+                x4_MicaBasedSheets,
```
eigenraven, line 94: These would be some of my ideas:
```
import static coremod.dsl.RecipeDsl.*; // put DSL utilities here
// Make these final, sorted alphabetically and implicitly x1
final ItemStack obsidianIngot = GT_OreDictUnificator.get(OrePrefixes.ingot, Materials.Obsidian, 2L);
// ...
GT_Values.RA.addXRecipe(
  items(4, obsidianIngot), // Comes from the RecipeDsl static import
  mB(144, lava), // make units clear, this would also let us add helpers like fluidIngots, fluidBuckets etc. for clearer units
  seconds(20), // 20 * SECONDS is also good, this would by a style choice
  voltAmps(TierEU.LV, 0.5f)
);
```
 - Make all the locals `final` if possible for sure
 - Add a static-importable DSL (domain-specific language) class with helper static methods and constants to keep the style consistent everywhere
 - Sort the local stacks alphabetically (just use the IDE/text editor sort lines functionality) to ensure a consistent order and clean diffs

src/main/java/com/dreammaster/gthandler/recipes/AlloySmelterRecipes.java
```diff
@@ -8,236 +11,343 @@
...
-import gregtech.api.enums.Materials;
-import gregtech.api.enums.OrePrefixes;
+import gregtech.api.enums.*;
 import gregtech.api.util.GT_ModHandler;
 import gregtech.api.util.GT_OreDictUnificator;
 import gregtech.api.util.GT_Utility;
 
 public class AlloySmelterRecipes implements Runnable {
 
+    final String modIDEnderIO = "EnderIO";
```
eigenraven, line 21: Mod ids should go into an enum somewhere

src/main/java/com/dreammaster/gthandler/recipes/AlloySmelterRecipes.java
```diff
@@ -8,236 +11,343 @@
...
     @Override
     public void run() {
+        registerCoreModRecipes();
+        registerEnderIORecipes();
+        registerTinkerConstructRecipes();
+
+    }
+
+    public void registerTinkerConstructRecipes() {
+        if (!Loader.isModLoaded(modIDTinkerConstruct)) {
```
eigenraven, line 33: If mod ids are in an enum, the enum could have a method for this like `Mods.TinkersConstruct.isLoaded()`

leagris: The `isLoaded` status could be cached to save the costly repeated calls to `Loader.isModLoaded`. Gregtech already caches some of those.

src/main/java/com/dreammaster/gthandler/recipes/AlloySmelterRecipes.java
```diff
@@ -8,236 +11,343 @@
...
+                x2_AluminoSilicateWools,
+                60 * SECONDS,
+                (int) (TierEU.RECIPE_LV / 2));
+
+        GT_Values.RA.addAlloySmelterRecipe(
+                x1_SapphireDust,
+                x1_SiliconDioxideDust,
+                x2_AluminoSilicateWools,
+                60 * SECONDS,
+                (int) (TierEU.RECIPE_LV / 2));
```
combusterf, line 112: I think you can make this easier and remove the risk of accidents from integer maths by having a method that takes `tier` and `amps` and returns an appropriate EU/t value. 

E.g. `TierEU.RECIPE_LV.amps(0.5)` and since there are plenty of recipes that use one specific fraction, TierEU.RECIPE_LV.almostOneAmp()` to get 30EU/t and its tiered derivatives.

src/main/java/com/dreammaster/gthandler/recipes/AlloySmelterRecipes.java
```diff
@@ -8,236 +11,343 @@
...
+        ItemStack x3_IridiumAlloyItemCasings = com.dreammaster.item.ItemList.IridiumAlloyItemCasing.getIS()
+                .splitStack(3);
+
+        ItemStack x1_Clay = new ItemStack(Items.clay_ball);
+
+        ItemStack x2_CokeOvenBricks = com.dreammaster.item.ItemList.CokeOvenBrick.getIS().splitStack(2);
+
+        ItemStack x1_ObsidianBlock = new ItemStack(Blocks.obsidian, 1, 0);
+
+        ItemStack x1_ObsidianIngot = GT_OreDictUnificator.get(OrePrefixes.ingot, Materials.Obsidian, 2L);
```
combusterf, line 88: I'd remove (most of) the empty lines

src/main/java/com/dreammaster/gthandler/recipes/AlloySmelterRecipes.java
```diff
@@ -8,236 +11,343 @@
...
+
+        ItemStack x1_SapphireDust = GT_OreDictUnificator.get(OrePrefixes.dust, Materials.Sapphire, 1L);
+
+        ItemStack x1_GreenSapphire = GT_OreDictUnificator.get(OrePrefixes.dust, Materials.GreenSapphire, 1L);
+
+        ItemStack x1_NetherQuartzDust = GT_OreDictUnificator.get(OrePrefixes.dust, Materials.GreenSapphire, 1L);
+
+        ItemStack x1_CertusQuartzDust = GT_OreDictUnificator.get(OrePrefixes.dust, Materials.CertusQuartz, 1L);
+
+        ItemStack x2_AluminoSilicateWools = CustomItemList.AluminoSilicateWool.get(2L);
```
combusterf, line 69: I'd drop the -s suffix for plurals, it's error prone when copy-pasting x1 stuff

src/main/java/com/dreammaster/gthandler/recipes/AlloySmelterRecipes.java
boubou19: `TierEU.LV` is 32 EU/t, while `TierEU.RECIPE_LV` is 30 EU/t. I'd be in favor to unify them, as the confusion can quickly be made

src/main/java/com/dreammaster/gthandler/recipes/AlloySmelterRecipes.java
boubou19: sure thing

src/main/java/com/dreammaster/gthandler/recipes/AlloySmelterRecipes.java
boubou19: agreed

src/main/java/com/dreammaster/gthandler/recipes/AlloySmelterRecipes.java
boubou19: I made a new enum for that now :)

src/main/java/com/dreammaster/gthandler/recipes/AlloySmelterRecipes.java
boubou19: I added a new enum for that, with some lazy cache population.

src/main/java/com/dreammaster/gthandler/ModIDs.java
```diff
@@ -0,0 +1,26 @@
...
+import cpw.mods.fml.common.Loader;
+
+public enum ModIDs {
+
+    EnderIO("EnderIO"),
+    TinkerConstruct("TConstruct");
+
+    public final String modID;
+
+    private final Map<String, Boolean> modsLoaded = new HashMap<>();
```
eigenraven, line 15: No need for a map here, just make it a `private Boolean modLoaded = null;`, there's one per instance. If a map was needed, it should probably be static in that case (and maybe concurrent)

boubou19: i removed the map. Is it ok like that? I just don't know when the enum constructor will be evaluated at runtime, so dunno if what i did in the constructor will work or not.

eigenraven: All constructors run during class loading, so I'd probably make it a Boolean and change it from null to a value on first use instead

boubou19: should be good now

PR #532, opened 2023-02-23
Chang drill head shape recipe

Ethryan:
Don’t think it conflicts.


Dream-Master:
@Ethryan why you change it? Just a visual thing ?

Ethryan:
It conflicts with the Coinage Mold. since that one uses the same spots for file and Wirecutter.
image

Dream-Master (approved):

PR #536, opened 2023-03-06
Iridium Neutron Reflector recipes skip

RaitGamerGR:
Added high-tier 1 step recipe for Iridium Neutron Reflector.
This recipe changed to avoid unstackable items on recipes. With any bus except for stocking old recipes works very slow because of 1 recipe use ~4 slots. With super buses you cant use all parallel of your assembler for this craft.


Dream-Master (approved):

Dream-Master:
Yea I wasn’t sure too

PR #538, opened 2023-03-10
Fix for Tungstensteel Ingots Needing TPV Coils

Steelux8:
- Corrected the appropriate recipes in order to set EBF temperature to 4000K, as intended by the last change.

See https://github.com/GTNewHorizons/GT5-Unofficial/pull/1628 for the PR that was originally meant to change this.


chochem (approved):
yep, this one works correctly now.

PR #547, opened 2023-03-27
Adjust according to tier

Runakai1:
Since the Industrial Apiary is in HV now, I changed the upgrades needing HV materials too. This was supposed to be done with the downgrade of the industrial Apiary, but I forgot. 

The Osmium and Palladium Gears were changed to steel and stainless steel gears.


kuba6000 (approved):

Dream-Master:
This is a huge buff hope you sure with tier.

PR #566, opened 2023-04-20
Fix materials

Dream-Master:
Need this pr

* https://github.com/GTNewHorizons/GT5-Unofficial/pull/1893


Phineasor (approved):
Needed for Steller cat, and material fix. Now has free thaumcraft fixes to make Good Gen build.

PR #567, opened 2023-04-22
Fix crash

kuba6000:


chochem (approved):
looks fine. also tested it and it does indeed stop the crashing.

mitchej123 (approved):

PR #570, opened 2023-04-22
Add Stellar Catalyst DTPF recipes

GDCloudstrike:
Adds a new set of dptf recipes for some materials using the new stellar catalyst. These new recipes mostly follow the pattern of the already existing ones, except for the transcendent metal recipe which is new to the dtpf.

This new recipe combines all the materials needed for tesseract -> transcendent metal conversion and condenses them into a dtpf recipe. There is an added material in the form of cosmic neutronium blocks to give the recipe a bit of an additional cost and make it a tad more interesting.

The recipe:
image

Also adjusted the catalyst values in the infinity recipes as those had slightly wrong values


BlueWeabo (approved):
i will trust you on this

PR #575, opened 2023-04-26
Move everything from the scripts to Core Mod

kuba6000:
Should be done in a few days/weeks
Scripts removal pr: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/13354

Script file (.zs) translators to code:
https://github.com/GTNewHorizons/CraftTweaker/commit/7fe4150afc7fa8d52799b45d6790f27ae9829e86
https://github.com/GTNewHorizons/ModTweaker/commit/25ae719ab98d2942ae013f16dcae2fc1c22cd0f6
https://github.com/GTNewHorizons/Minetweaker-Gregtech-5-Addon/commit/625cd00f385254d2732985ac4e57a3903391fcfa
https://github.com/GTNewHorizons/BloodMagic/commit/4c9c955e6310a82d0ef0a8f4f7963999b3ce062e
https://github.com/GTNewHorizons/Avaritia/commit/c1e4f5e235af43be05e31cbaa458febfd7baa869

TODO (look here: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/tree/scripts-are-gone/scripts):
- [x] Crafting recipes
- [x] Furnace recipes
- [x] All mods.\<something\>....
- [x] Name overrides
- [x] Hidden items
- [x] Tool tips
- [x] ZZ-Client-Only
- [x] Dependencies


Chaoschaot232:
I must say great work @kuba6000 !
This is one of the best improvments made to the pack speaking of loading times and server login times 👍

Keep up the good work 👍😊

boubou19:
@kuba6000 a few remarks on this drafted PR:
- the zz_client_only.zs should remain as a script, because i know a lot of players who use this in multiplayer. Also the way it is currently planned is making them availiable for MP too, where Dream explicitely wants everything in this file to be SP only.
- the dependencies are totally wrong: many calls to `getModItem` are done with other mods than the ones declared in dependencies, which may lead to unpredictible behavior if one of the required mods is missing. Please correct.
- all the mod ids are hardcoded, use the enum from GT5U ( make a PR to GT5U if a mod id is missing). ctrl+shift+r and ctrl+shift+f are your friends if you use Idea with default keybinds. (search for `getModItem("` with extended search to find every single hardcoded mod id usage)
- The usage of a placeholder for missing stuff shouldn't be required at all, as you are supposed to declare all the depedencies to the script: either all the recipes can be loaded, or none. Main goal of those recipes is to be added in full pack. if one or more mods are missing, then it's either dev env, in which case you want to disable all recipes to gain loading time, or players have an incorrect setup (unsupported). To remove all the placeholders, just use ctrl+shift+r and search for `, missing)` and replace it by `)`
- wildcard imports are bad: those can cause weird stuff, so best is to avoid all of its usage.[ Here](https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports) is how to turn them off. Also, to fix all of them in the project, just right click on the `java` folder, and click on `optimize imports` (uncheck `process only VCS changed files` to eventually catch all wildcard imports which made in the codebase). You'll get tons of changes to commit, but rerun spotless first, it will remove a ton of changes.

kuba6000:
> @kuba6000 a few remarks on this drafted PR:
> - the zz_client_only.zs should remain as a script, because i know a lot of players who use this in multiplayer. Also the way it is currently planned is making them availiable for MP too, where Dream explicitely wants everything in this file to be SP only.
> - the dependencies are totally wrong: many calls to `getModItem` are done with other mods than the ones declared in dependencies, which may lead to unpredictible behavior if one of the required mods is missing. Please correct.
> - all the mod ids are hardcoded, use the enum from GT5U ( make a PR to GT5U if a mod id is missing). ctrl+shift+r and ctrl+shift+f are your friends if you use Idea with default keybinds. (search for `getModItem("` with extended search to find every single hardcoded mod id usage)
> - The usage of a placeholder for missing stuff shouldn't be required at all, as you are supposed to declare all the depedencies to the script: either all the recipes can be loaded, or none. Main goal of those recipes is to be added in full pack. if one or more mods are missing, then it's either dev env, in which case you want to disable all recipes to gain loading time, or players have an incorrect setup (unsupported). To remove all the placeholders, just use ctrl+shift+r and search for `, missing)` and replace it by `)`
> - wildcard imports are bad: those can cause weird stuff, so best is to avoid all of its usage.[ Here](https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports) is how to turn them off. Also, to fix all of them in the project, just right click on the `java` folder, and click on `optimize imports` (uncheck `process only VCS changed files` to eventually catch all wildcard imports which made in the codebase). You'll get tons of changes to commit, but rerun spotless first, it will remove a ton of changes.

I want to make that zz-client-only load on single player and an config option to also load on server.

I dont look at dependencies currently at all. I want to do them when scripts folder will be empty

I dont think that the missing placeholder is bad at all. If for some unknown reason, an item will dissapear then it will be displayed in the game as a fire with custom name to report in on github.

Personally, I like the wildcard imports, but if you really want to remove them I can do that at the end.

boubou19:
ok for 1 and 2, 3 can be discussed, but the [accepted answer](https://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad) of this stackoverflow post explain why wildcard imports are bad. I thought it would never occur on any NH project i work on, but one day it happened to me in coremod. Now i yeet them.

Dream-Master:
lets test on zeta

Dream-Master (approved):

PR #598, opened 2023-05-13
Fix recipe remover and some recipe additions

kuba6000:
```
[09:55:04] [Client thread/INFO] [DREAMCRAFT/dreamcraft]: Removed 4254 recipes!
[09:55:04] [Client thread/INFO] [DREAMCRAFT/dreamcraft]: Recipes removal took 42 ms.
```
Fixes  https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/13492
Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/13494


chochem:
not a code review. just loading it up and comparing it to the ticket. I still get these two:
image
image

chochem:
thanks, those two look good now too

wlhlm:
The wildcard meta value (32767), this that something specific to how GTNH Core Mod works, or is this a special value from RA? If the latter, I think we should add it as a proper constant in RA rather than have it as a variable in an interface here.

kuba6000:
> The wildcard meta value (32767), this that something specific to how GTNH Core Mod works, or is this a special value from RA? If the latter, I think we should add it as a proper constant in RA rather than have it as a variable in an interface here.

It is just normal wildcard meta, also defined as a static constant variable in OreDictionary class I think. Nothing to do with RA

OneEyeMaker:
Why don't you use the constant (`OreDictionary.WILDCARD_VALUE`) instead of "magic" number?

kuba6000:
> Why don't you use the constant (`OreDictionary.WILDCARD_VALUE`) instead of "magic" number?

What do you mean? These recipe removals were ported straight from scripts and I just replaced all the magic wildcard numbers with a variable in IScriptLoader class named "wildcard" or did I miss something?

src/main/java/com/dreammaster/scripts/ScriptAdvancedSolarPanel.java
```diff
@@ -31,65 +31,91 @@ public void loadRecipes() {
...
-                        getModItem(AdvancedSolarPanel.ID, "BlockAdvSolarPanel", 1, 0, missing),
-                        "craftingToolScrewdriver", "circuitAdvanced",
-                        getModItem(IndustrialCraft2.ID, "itemArmorNanoHelmet", 1, 32767, missing), "circuitAdvanced",
-                        "cableGt04Aluminium", getModItem(GregTech.ID, "gt.blockmachines", 1, 22, missing),
-                        "cableGt04Aluminium" });
+                "craftingToolWrench",
+                getModItem(AdvancedSolarPanel.ID, "BlockAdvSolarPanel", 1, 0, missing),
+                "craftingToolScrewdriver",
+                "circuitAdvanced",
+                getModItem(IndustrialCraft2.ID, "itemArmorNanoHelmet", 1, 32767, missing),
```
OneEyeMaker, line 38: This one and two more in this file.

src/main/java/com/dreammaster/scripts/ScriptAvaritiaAddons.java
```diff
@@ -78,17 +89,21 @@ public void loadRecipes() {
...
+                getModItem(GregTech.ID, "gt.metaitem.01", 1, 32633),
+                "circuitElite",
+                getModItem(GregTech.ID, "gt.metaitem.01", 1, 32633));
 
         // clearing NBT of th
-        addShapelessCraftingRecipe(
+        addShapelessRecipe(
                 getModItem(AvaritiaAddons.ID, "ExtremeAutoCrafter", 1),
-                new Object[] { getModItem(AvaritiaAddons.ID, "ExtremeAutoCrafter", 1, 32767), "platePaper" });
+                getModItem(AvaritiaAddons.ID, "ExtremeAutoCrafter", 1, 32767),
```
OneEyeMaker, line 105: This one.

src/main/java/com/dreammaster/scripts/ScriptAdvancedSolarPanel.java
kuba6000: oh ye i just replaced these in RecipeRemovals file

Dream-Master (approved):
works ingame so far i can see ingame

PR #599, opened 2023-05-15
Remove Universal Fluid Cell from recipes

miozune:


src/main/java/com/dreammaster/scripts/ScriptIndustrialCraft.java
```diff
@@ -1330,17 +1331,6 @@ public void loadRecipes() {
...
-                createItemStack(
-                        IndustrialCraft2.ID,
-                        "itemFluidCell",
-                        1,
-                        0,
-                        "{Fluid:{FluidName:\"ic2biogas\",Amount:1000}}",
-                        missing),
-                getModItem(IndustrialCraft2.ID, "itemFluidCell", 1, 0, missing),
-                40,
-                1);
```
OneEyeMaker, line 1343: Is removal of biogas as fuel intended?

miozune: `addFuel` with `aType` 1 means adding to `sTurbineFuels`. And we already have biogas turbine fuel recipe, added by GT.

OneEyeMaker (approved):

PR #600, opened 2023-05-15
fix laser engraver consuming the lens

Fiyorin:
the lens isn't marked "NC" and gets consumed in the crafting
this fixes it

image

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/LaserEngraverRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/LaserEngraverRecipes.java
@@ -255,28 +255,28 @@ public class LaserEngraverRecipes implements Runnable {
         // Optical to Wafer Buff Recipes
         GT_Values.RA.addLaserEngraverRecipe(
                 ItemList.Circuit_Silicon_Wafer6.get(1L),
-                GT_OreDictUnificator.get(OrePrefixes.lens, Materials.NetherStar, 1L).copy().splitStack(0),
+                GT_Utility.copyAmount(0, GT_OreDictUnificator.get(OrePrefixes.lens, Materials.NetherStar, 1)),
                 CustomItemList.RawPicoWafer.get(16),
                 6000,
                 1887440,
                 true);
         GT_Values.RA.addLaserEngraverRecipe(
                 ItemList.Circuit_Silicon_Wafer6.get(1L),
-                GT_OreDictUnificator.get(OrePrefixes.lens, Materials.Emerald, 1L).copy().splitStack(0),
+                GT_Utility.copyAmount(0, GT_OreDictUnificator.get(OrePrefixes.lens, Materials.Emerald, 1)),
                 ItemList.Circuit_Wafer_SoC2.get(32),
                 45 * 20,
                 122880,
                 true);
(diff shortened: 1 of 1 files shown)


src/main/java/com/dreammaster/gthandler/recipes/LaserEngraverRecipes.java
```diff
@@ -255,28 +255,28 @@ public void run() {
         // Optical to Wafer Buff Recipes
         GT_Values.RA.addLaserEngraverRecipe(
                 ItemList.Circuit_Silicon_Wafer6.get(1L),
-                GT_OreDictUnificator.get(OrePrefixes.lens, Materials.NetherStar, 1L).copy(),
+                GT_OreDictUnificator.get(OrePrefixes.lens, Materials.NetherStar, 1L).copy().splitStack(0),
```
Dream-Master, line 258: why not set 1L to 0L ?
like in this recipe https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/600/files#diff-99457ece25a9877f6a71a1e17dffcebe33cc56c5d36d69b707b515bfbce3edcfL219

Fiyorin: i don't know, @TheEpicGamer274  wrote the code

minecraft7771: Use 0L as dream suggested.

chochem: GT_OreDictUnificator would just give null

chochem: you could use GT_Utility.copyAmount(0, GT_OreDictUnificator.get(OrePrefixes.lens, Materials.NetherStar, 1)) though to keep it consistent with the rest of the file

minecraft7771: Hmm, we should change that possibly or what do you think about it?

TheEpicGamer274: I agree with chochem

Fiyorin:
> sorry about the kerfuffle bois

no problem, this 100x better than to continue using the old recipe

src/main/java/com/dreammaster/gthandler/recipes/LaserEngraverRecipes.java
Fiyorin: ye i'm testing it with chochems idea

Fiyorin: it would just crash if i gave it 0L

TheEpicGamer274: sounds about right

TheEpicGamer274: I believe my water line recipes have the exact thing for lens use that is the best to use usually

chochem (approved):

PR #615, opened 2023-05-19
update tank capacity on GT machines in coremod

Pxx500:
second part of https://github.com/GTNewHorizons/GT5-Unofficial/pull/1999


Dream-Master:
Why 128 is max. Did I miss a vote or a suggestion about this value ?

Pxx500:
> Why 128 buckets is max. Did I miss a vote or a suggestion about this value ?

Anything higher than 128 makes no sense since most of the fluid will be idling inside the tank and not buffering since pretty much no recipes use more than a few buckets anymore after chemical rebalance. Also all machines in coremod are high-tier and most will be used in PA anyway so fluid tank capacity is irrelevant for the most part. I used 128 because it seems like a good compromise between being low enough to not seem weird and high enough to not cause buffering problems.

Dream-Master:
ok fair enough but then the gt pr is wrong. it set a default to 64 not 128. Andy why you not use the code from gt instead of type in the values?

Pxx500:
> ok fair enough but then the gt pr is wrong. it set a default to 64 not 128. Andy why you not use the code from gt instead of type in the values?

Not that i think about it 256 is also very convenient since drums have this value. Maybe we'll go with that?

Dream-Master (approved):

PR #626, opened 2023-05-25
Fix obsidian stick conversion recipes

chochem:
They are conflicting and which one you actually get seems to depend on version/restart/or something.

I made them shapedore recipes instead. topleft you get the tinker version, topmiddle randomthings, topright GT. Any works as input.

Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10011.

image

edit: uh something still isnt right. draft until I fix it (or someone else). Right now the first 2 are working but I also get the tinker variant topright even though it shows gt in nei. :( Looks like the recipe conflict was just one of the problems here.
edit2:restart and now its wrong the otherway around. topmiddle and topright are correct but topleft gives GT. uff. will try to understand on the weekend.


EnderProyects:
Why you dont make a recipe that cycles the 3 recipes instead of making a place one?

chochem:
I dont see how that is relevant. seems better this way.

EnderProyects:
Because it will not cause conflicts

chochem:
alright kuba and boubou quickly helped. problem was that it's not position but really a 'shape'. here are the new ones, all working now:
image

PR #629, opened 2023-05-30
Clean up the minecraft script

chochem:
- use SECONDS and other time units
- use TierEU for voltages
- use customitemlist for coremod items
- use itemlist for gt items
- use gt_oredictunificator for gt materials
- use material.x.getfluid/gas/molten for gt fluids
- remove duplicate recipes

overall much more readable and much fewer gameregistry/fluidregistry queries

tested in game. all seems to work


wlhlm (approved):
Nice work!

PR #631, opened 2023-06-01
Add Grate Machine Casing assembler recipe 

DotJason:
Addresses [#13571](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/13571)


chochem:
this method of adding recipes is deprecated. please use RA2. see https://github.com/GTNewHorizons/GT5-Unofficial/blob/master/src/main/java/gregtech/loaders/postload/recipes/AssemblerRecipes.java for lots of assembler examples.

DotJason:
Thanks for the heads up, I'll rewrite it

PR #471, opened 2022-12-30
Add recipe for Hyper-Acceleration Card

S4mpsa:
Gated at ZPM with a single Mysterious Crystal per card.

https://github.com/GTNewHorizons/Applied-Energistics-2-Unofficial/pull/218

image

PR #633, opened 2023-06-02
Two recipe improvements

chochem:
- add a _proper_ recipe for the hyperacceleration card. this was left "for later" here https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/471#issuecomment-1368061519
- 'fix' fluid amounts for alveary blocks in the new assembler recipes. They are now the same as in the old and still existing carpenter recipes. While the changed amounts might have been intentional, there was no mention of a balance change in the pr and no such discussion, nor does it make any sense to have different amounts in different machines. They were also inconsistent between the blocks, at least since https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/623.

helpful pic:
image

### DIFF
--- a/src/main/java/com/dreammaster/scripts/ScriptAppliedEnergistics2.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptAppliedEnergistics2.java
@@ -5,7 +5,6 @@ import static gregtech.api.enums.Mods.Avaritia;
 import static gregtech.api.enums.Mods.BartWorks;
 import static gregtech.api.enums.Mods.EternalSingularity;
 import static gregtech.api.enums.Mods.ExtraUtilities;
-import static gregtech.api.enums.Mods.GTPlusPlus;
 import static gregtech.api.enums.Mods.GregTech;
 import static gregtech.api.enums.Mods.IndustrialCraft2;
 import static gregtech.api.enums.Mods.IronChests;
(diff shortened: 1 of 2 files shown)


src/main/java/com/dreammaster/scripts/ScriptAppliedEnergistics2.java
```diff
@@ -62,7 +63,8 @@ public List<String> getDependencies() {
                 ExtraUtilities.ID,
                 IndustrialCraft2.ID,
                 IronChests.ID,
-                Thaumcraft.ID);
+                Thaumcraft.ID,
+                GTPlusPlus.ID);
```
kuba6000, line 67: Where is this dep needed?

chochem: while it is a gt.blockmachines, it comes from gtpp

kuba6000: Which one? If thats an world accelerator then they are in core mod

chochem: damn you are right, thanks for checking. my bad.

chochem: fixed

kuba6000 (approved):
Looks fine

PR #634, opened 2023-06-04
Improve EMT charging ring recipes

chochem:
closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/13611

In short, this brings the EMT Inventory and Armor Charging Rings down from EV to HV. Thus making them more comparible with other options like the nanosolarhelmet or LV solar(or other power) + LV wireless.
(was also brought to meta-dev with no disagreements sofar but some support)

new recipes as discussed in the ticket:
image
image


Dream-Master (approved):

Alastors:
nice

PR #643, opened 2023-06-13
More qol assembler + extruder recipes + fix assembling line eu cost

TheEpicGamer274:
More QOL assembler recipes to further reduce molecular assembler requirements in crafting.


chochem (approved):
looks good, thanks a lot!

PR #645, opened 2023-06-14
new hempcrete recipes

chochem:
Adds new recipes for hempcrete. goes together with https://github.com/GTNewHorizons/Chisel/pull/30

image
image

closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/13702

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/GT_CraftingRecipeLoader.java
+++ b/src/main/java/com/dreammaster/gthandler/GT_CraftingRecipeLoader.java
@@ -12,7 +12,15 @@ import com.dreammaster.main.NHItems;
 import com.github.bartimaeusnek.bartworks.common.loaders.ItemRegistry;
 import com.github.bartimaeusnek.bartworks.system.material.WerkstoffLoader;
 
-import gregtech.api.enums.*;
+import gregtech.api.enums.Dyes;
+import gregtech.api.enums.GT_Values;
+import gregtech.api.enums.ItemList;
+import gregtech.api.enums.Materials;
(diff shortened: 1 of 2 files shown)


Dream-Master:
why not give a discount on assembler recipes ?

chochem:
can do yea. just higher output?

Dream-Master (approved):

Darkmaxsas:
@chochem is there a reason it now requires 1 Raw Bio Fibre per 12 Blocks Hempcrete?
Seems kinda expensive (thats like 2 stacks saplings/other plant materials for 12 Hempcrete powder)
or is the recipe for raw bio fibre changed as well?

boubou19:
it's likely to change again, to plant balls

chochem:
@Darkmaxsas The main critique points with the old recipe were the weird duo of 1) its very simplistic 2) its terrible to automate (you need botania).

this now is the opposite, its not simplistic but fully automatable with lv machines.

That makes it now an interesting goal to go for if you want to automate/massproduce it in LV. And then it becomes rather easy in MV and of course is trivial for anyone in higher tiers.

Darkmaxsas:
yeah i see your point the original recipe isn't good either. But i think the cost now is a tad too high for the Block to be usable at large scales compared to other chisel/zTones blocks. Maybe change the raw bio fibre to plantballs like boubou mentioned.

And move the recipe for the final block from just "smelting" the powder to running the powder through an autoclave with water/distilled water.

That way you keep the goal of automating it and not dropping the complexity.

PR #646, opened 2023-06-15
coremod bending recipe cleanup

chochem:
- moves all the bending recipes from scripts to the bending file
- remove duplicate recipes that never were in the game as they were replaced by the automatically generated one
- readded important circuits to recipes that got lost when script files were moved to code
- replaced ids with oredictunificator
- **added one new recipe**: 9 obsidian ingot -> 1 dense plate (it was always the only material that didnt have this)
- use tiereu and seconds/minutes

fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/13693


TheEpicGamer274 (approved):
Looks good.

PR #650, opened 2023-06-19
Fix oredict problems in LP recipes

wlhlm:
Fixes the following:

LP recipe bug

Also refactors the LP recipes to use constants for power usage and duration.


wlhlm:
> I dont know if you want to get into that, but ideally getmoditem should not be used for GT and coremod items.

I'm not really interested at this time. I'm not too familiar with all the recipe handling, just wanted to fix a small thing I've hit in my play hrough.

PR #653, opened 2023-06-22
ChemBath RA2 + cleanup

chochem:
Converted the Chemical Bath recipes to RA2, collected all the chem bath recipes from the script files, and cleaned up (use SECONDS and TierEU.RECIPE, get GT items and fluids directly instead of from registry, fix some modloaded checks, etc.).

I checked a few selected recipes in-game to make sure I didnt break things, but not every one of them.


Steelux8 (approved):
I'll trust the testing, thank you for the RA2 switch and organization.

PR #654, opened 2023-06-28
Assembling Machine recipes for Lootbag upgrading and downgrading

Cleptomania:
This adds recipes as discussed in https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/11861

In general, these are gated by the same tier of assembler that the questbook would gate them by. For example, you need an MV assembler to turn LV lootbags into MV, and an EV to turn HV in to EV.

This is also loosely followed for the other various types of lootbags. For example, the first tier of forestry upgrades in the quest book can be accessed during LV, however you need at least MV tier power/components to get to the next tier of forestry lootbag upgrades based on the quests required, so you need an MV assembler to do that upgrade.

All of these recipes currently use a full amp of power for their respective tier, and take 30 seconds to complete. This of course can be adjusted if it should be different, I just kind of threw these numbers in there.

This also allows downgrading into the previous tier of lootbags via the assembler. Upgrading/downgrading is controlled via circuit configuration.


Dream-Master (dismissed):
I am against it. You need no additional item or a enchanted book.

Dream-Master:
@Cleptomania ok if you choose only not enchanted lootbags this is ok. But please change the recipe adder to version 2

Steelux8 (changes requested):
The ability to downgrade lootbags needs some balance-related discussion, regardless of how it's done. Even with just GT tier lootbags, there's already a problem where tiers above LuV would end up rewarding dozens of IV lootbags, and a huge supply of free ender chests, tanks and other stuff. The impact of IV lootbags is already significant enough as it is, and I think we need to talk about whether or not we want to give even more of that free stuff by allowing lootbag downgrades.

Cleptomania:
Perhaps downgrading can be done at a 1-1 ratio instead of giving back the original 3, such that you can get a lower tier loot bag if you need but you can’t turn higher tier rewards into more loot bags than you originally got.

Cleptomania:
I've switched the recipes to use the new builder, as well as went ahead and removed the downgrade recipes, at least until those can be discussed further, so for now it only allows upgrades exactly as the questbook currently does.

@Dream-Master I didn't realize when I made this that the old `addAssemblerRecipe` and other similar functions were deprecated. Is there are a desire to remove all usage of those from this? Would you want a PR for updating all of the recipes still using that API?

chochem:
yea, we arein the process of transforming all those to the new RA2. GT is done. coremode like half-way maybe.

chochem (changes requested):
just some more technical points:
- why is this placed in a completely unrelated script file? this should be in the assembler recipe file and with a ismodloaded condition for EnhancedLootbags around it.
- dont use fullamps, people have loss. use TierEU.RECIPE_MV and similar (that is easy to read and cuts off 1/16th)
- use 30*SECONDS instead 600 (our devs dont like unclear magic numbers as you can tell)

Cleptomania:
I added new changes with the feedback from @chochem

I agree that the AssemblerRecipes is the most sensible place for it, the organization of stuff just wasn't initially clear to me.

I can follow up this PR with a full cleanup of Assembler recipes with the more up to date APIs and conventions(thanks Chochem your previous PR for the chembath stuff is a good frame of reference for that) if no one else is already working on that, but this PR should probably stay separate since it's actually adding new recipes I'd imagine.

Dream-Master:
@Steelux8

chochem (approved):

Dream-Master (approved):

Cleptomania:
I saw you pulled in my `assembler-cleanup` cleanup branch for RA2 conversion into the `dev` branch. Assuming that’s going to master at some point you could close this PR as that already had these changes applied to it.

wlhlm:
The dev branch is for making -pre releases for testing changes in nightlies. Accepted changes land on master/main.

chochem:
you will need to make a PR for that. and apparently it has some issues and needs some splitting to work.

PR #659, opened 2023-07-03
Initial Method Breakout/Organization for AssemblerRecipes

Cleptomania:
This is a preliminary PR to a bunch of the RA2 work that needs to be done on AssemblerRecipes.

In testing RA2 conversions, it was found that the `run()` method had become too large and needed trimmed apart into more functions. In addition to needing done, doing this work first will make it easier to submit smaller, more isolated and easier to review PRs for the upcoming RA2 conversions.

This PR does not modify any recipes, and should not change any behavior. The only thing done in this PR is organizing many of recipes into their own functions grouped either by the mod they are for, or their purpose as best can be defined. There are still a good amount of recipes in the primary `run()` method, however they are ones that don't have many of them to group into a function, so these can probably be tackled later.

Once this is merged, it will open the door to doing RA2 conversions on the existing methods one at a time in a much easier way for maintainers to deal with hopefully.


chochem (approved):
looks fine from what I can tell.

Dream-Master:
Recipes shown ingame. i tested a few look good

PR #660, opened 2023-07-04
Fix furnace smelting recipes from script conversion

kuba6000:
Correctly mimic what MineTweaker was doing
https://github.com/GTNewHorizons/CraftTweaker/blob/03ca9b2055215ebce96ab0ccc4d67c2b48b2091c/src/main/java/minetweaker/mc1710/furnace/MCFurnaceManager.java#L163
Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/13826


Dream-Master (approved):

Dream-Master:
works
image

chochem:
hmm. basically gtmodhandler does the exact same thing. the only difference is that it checks for the input having a containeritem first and does not add a recipe then. Which is indeed the case for the genesample (the container being the blank gene sample). All other recipes were already working.

Is there any good reason for gtmodhandler doing that check? if not then maybe it would be cleaner to just remove that instead of calling these unnamed MC functions directly each time.

context is here: https://github.com/GTNewHorizons/GT5-Unofficial/blob/8115282baedea5d663ef2a8fb40091804de32e86/src/main/java/gregtech/api/util/GT_ModHandler.java#L529C17-L529C17

in particular the `GT_Utility.getContainerItem(aInput, false) != null` ->return false

chochem:
so am I wrong or just being ignored?

fluffle:
The research I dropped in the linked issue shows that the container check is stopping about a half dozen "cook a cell of molten metal into its ingot" recipes from being added. So it is doing *something*, albeit not all that much. Arguably not trying to create those recipes in the first place would be better but it's probably worth having a check.

It is also evident that of all the recipes @kuba6000 changed, only the GeneSample one *needed* to be changed. The others could safely have been left as-is. But I suspect what he did was change his "generate Java to match ZenScript" code such that it just output the direct furnace recipe addition and re-run it, rather than manually edit all of these files...

chochem:
ah yea that lines up with what I said, thanks for testing out what is affected (any of those would be fine either way I would say).

I would say we rather undo this and change the method. cant expect people to remember func_151394_a just because they want a stupid furnace recipe. func_151394_a also doesnt exactly make these recipes clear or easy to read. But I dont know, seems dream disagrees or something.

Dream-Master:
Kuba make a new pr dont worry

PR #662, opened 2023-07-05
Time <-> Space conversion recipes

GDCloudstrike:
Adds conversion recipe for time to space and vice versa.

Molten space and molten time are sourced from the same recipe which yields both, and a lot of times one ends up with too much of either one. Plus the eternity qft recipe yields excess molten time which just sits there because you also get it when crafting molten space. This is especially noticeable very late into the game when primordial matter becomes a relevant crafting ingredient and molten space usage becomes higher than molten time usage.

These recipes aim to address these issues by adding a way to convert the two fluids into one another.

Time to Space:
image

Space to Time:
image


Connor-Colenso (approved):

Dream-Master:
@GDCloudstrike feel free to merge.

PR #664, opened 2023-07-05
Add compressor recipe for nether star block from dust

greesyB:
Gives nether star dust a use and alleviates some of the extreme endgame need for nether stars/plates

image

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/CompressorRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/CompressorRecipes.java
@@ -9,6 +9,8 @@ import static gregtech.api.enums.Mods.OpenComputers;
 import static gregtech.api.enums.Mods.StevesCarts2;
 import static gregtech.api.enums.Mods.Thaumcraft;
 import static gregtech.api.enums.Mods.TinkerConstruct;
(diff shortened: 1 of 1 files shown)


Dream-Master:
@greesyB isnt it better make a implosion compressor recipe make stars from dust and make a block from them?
Sure its a UV recipe but if you add it we need to add some for other stuff as well.

greesyB:
I don't think this recipe should be available too early, so if it is an implosion compressor recipe it should probably be limited to electric implosion only. Is that preferred?

Dream-Master:
i guess when you can make a electric implosion compressor its not needed anymore. Is there an recipe in the autoclave or such other ways make powder to a star/block before uv ?

Dream-Master:
nevermind there are at HV

Dream-Master (approved):

greesyB:
I added this recipe due to how many nether stars a stargate needs as the autoclave recipe is not viable to get 300+ million stars even with UMV processing arrays. I made it UV so it could have some use a little earlier, but I don't have a problem with making it an electric implosion recipe instead or even a normal implosion recipe if it is wanted for early game.

GDCloudstrike (approved):

PR #668, opened 2023-07-06
Convert NEI Planet Recipes to RA2

Cleptomania:
Converts all of the assembler recipes for the NEI Ore Plugin dimension blocks to RA2.

All checks are also put behind applicable `isModEnabled` checks. If `GalaxySpace` is available then it is assumed `GalacticraftCore` and `GalacticraftMars` are available.

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
@@ -23,7 +23,10 @@ import com.dreammaster.gthandler.CustomItemList;
 import com.github.bartimaeusnek.bartworks.common.loaders.ItemRegistry;
 
 import gregtech.api.enums.*;
-import gregtech.api.util.*;
+import gregtech.api.util.GT_ModHandler;
+import gregtech.api.util.GT_OreDictUnificator;
+import gregtech.api.util.GT_Recipe;
+import gregtech.api.util.GT_Utility;
 import gregtech.common.items.GT_MetaGenerated_Tool_01;
 import gtPlusPlus.core.recipe.common.CI;
 
@@ -7278,7 +7281,7 @@ public class AssemblerRecipes implements Runnable {
                         GT_Utility.getIntegratedCircuit(1))
                 .fluidInputs(Materials.Water.getFluid(10000))
                 .itemOutputs(getModItem(NEIOrePlugin.ID, "blockDimensionDisplay_Ow", 1, 0)).noFluidOutputs()
-                .duration(300).eut(TierEU.RECIPE_LV).addTo(sAssemblerRecipes);
+                .duration(15 * SECONDS).eut(TierEU.RECIPE_LV).addTo(sAssemblerRecipes);
         // Twilight Forest
         if (TwilightForest.isModLoaded()) {
             GT_Values.RA.stdBuilder()
@@ -7290,7 +7293,7 @@ public class AssemblerRecipes implements Runnable {
                             GT_Utility.getIntegratedCircuit(2))
                     .fluidInputs(Materials.Water.getFluid(10000))
                     .itemOutputs(getModItem(NEIOrePlugin.ID, "blockDimensionDisplay_TF", 1, 0)).noFluidOutputs()
-                    .duration(300).eut(TierEU.RECIPE_LV).addTo(sAssemblerRecipes);
+                    .duration(15 * SECONDS).eut(TierEU.RECIPE_LV).addTo(sAssemblerRecipes);
         }
         // Nether
         if (BiomesOPlenty.isModLoaded()) {
@@ -7303,7 +7306,7 @@ public class AssemblerRecipes implements Runnable {
                             GT_Utility.getIntegratedCircuit(3))
                     .fluidInputs(Materials.Lava.getFluid(10000))
                     .itemOutputs(getModItem(NEIOrePlugin.ID, "blockDimensionDisplay_Ne", 1, 0)).noFluidOutputs()
-                    .duration(300).eut(TierEU.RECIPE_LV).addTo(sAssemblerRecipes);
+                    .duration(15 * SECONDS).eut(TierEU.RECIPE_LV).addTo(sAssemblerRecipes);
         }
         // End
         if (Natura.isModLoaded()) {
@@ -7316,7 +7319,7 @@ public class AssemblerRecipes implements Runnable {
                             GT_Utility.getIntegratedCircuit(4))
                     .fluidInputs(FluidRegistry.getFluidStack("ender", 10000))
                     .itemOutputs(getModItem(NEIOrePlugin.ID, "blockDimensionDisplay_ED", 1, 0)).noFluidOutputs()
-                    .duration(300).eut(TierEU.RECIPE_LV).addTo(sAssemblerRecipes);
+                    .duration(15 * SECONDS).eut(TierEU.RECIPE_LV).addTo(sAssemblerRecipes);
         }
         // Far End Asteroids
         if (HardcoreEnderExpansion.isModLoaded()) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
```diff
@@ -23,10 +23,7 @@
 import com.github.bartimaeusnek.bartworks.common.loaders.ItemRegistry;
 
 import gregtech.api.enums.*;
-import gregtech.api.util.GT_ModHandler;
-import gregtech.api.util.GT_OreDictUnificator;
-import gregtech.api.util.GT_Recipe;
-import gregtech.api.util.GT_Utility;
+import gregtech.api.util.*;
```
OneEyeMaker, line 26: Please, replace wildcard import with regular ones.

OneEyeMaker, line 26: Please replace wildcart import with multiple regular ones.

src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
```diff
@@ -2121,6 +2118,24 @@ public void run() {
...
+        GT_Values.RA
+                .addAssemblerRecipe(
+                        new ItemStack[] {
+                                GT_OreDictUnificator
+                                        .get(OrePrefixes.pipeQuadruple, MaterialsUEVplus.TranscendentMetal, 1L),
+                                CustomItemList.Hull_UIV.get(1L), GT_Utility.getIntegratedCircuit(4) },
+                        Materials.Polybenzimidazole.getMolten(2304L),
+                        ItemList.Hatch_Input_Multi_2x2_UIV.get(1L),
+                        600,
+                        24);
```
OneEyeMaker, line 2137: Why these recipes don't use RA2?

src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
```diff
@@ -7252,428 +7267,422 @@ private void makeCoilRecipes() {
...
-                    FluidRegistry.getFluidStack("endergoo", 10000),
-                    GT_ModHandler.getModItem(NEIOrePlugin.ID, "blockDimensionDisplay_EA", 1L, 0),
-                    300,
-                    30);
+        // T0 Planets
+        // Overworld
+        GT_Values.RA.stdBuilder()
+                .itemInputs(
+                        getModItem(Botania.ID, "tinyPlanetBlock", 1, 0),
+                        getModItem(Minecraft.ID, "stone", 64, 0),
```
OneEyeMaker, line 7275: (Optionally) Vanilla items can be referenced directly, without `getModItem` (`new ItemStack(Blocks.stone, 64)`).

src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
```diff
@@ -7252,428 +7267,422 @@ private void makeCoilRecipes() {
...
+                        getModItem(Minecraft.ID, "grass", 64, 0),
+                        getModItem(Minecraft.ID, "sapling", 1, 0),
+                        GT_Utility.getIntegratedCircuit(1))
+                .fluidInputs(Materials.Water.getFluid(10000))
+                .itemOutputs(getModItem(NEIOrePlugin.ID, "blockDimensionDisplay_Ow", 1, 0)).noFluidOutputs()
+                .duration(300).eut(TierEU.RECIPE_LV).addTo(sAssemblerRecipes);
```
OneEyeMaker, line 7281: Use `SECONDS` (from `GT_RecipeBuilder`) for duration - makes recipes more readable.

OneEyeMaker (changes requested):

src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
chochem: the file is massive and the author didnt convert everything yet (nor did they claim to). those 2 just moved

src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
Cleptomania: Is there a preferred way to do this for vanilla items? To me the `getModItem` method is more clear, but if the norm is to just use the ItemStack constructor directly then I can do that.

src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
Cleptomania: This has been fixed now. I think IntelliJ did this automatically or something at some point.

src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
Cleptomania: Thanks for the catch, they're all using this now.

src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
OneEyeMaker: It's OK to do it both ways. Difference only is that `getModItem` searches for item before creating `ItemStack`.
And as I said, it's optional.

OneEyeMaker (approved):

src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
```diff
@@ -7252,428 +7270,422 @@ private void makeCoilRecipes() {
...
+                .duration(15 * SECONDS).eut(TierEU.RECIPE_LV).addTo(sAssemblerRecipes);
+        // Twilight Forest
+        if (TwilightForest.isModLoaded()) {
+            GT_Values.RA.stdBuilder()
+                    .itemInputs(
+                            getModItem(Botania.ID, "tinyPlanetBlock", 1, 0),
+                            getModItem(Minecraft.ID, "stone", 64, 0),
+                            getModItem(Minecraft.ID, "grass", 64, 0),
+                            getModItem(TwilightForest.ID, "tile.TFSapling", 1, 1),
+                            GT_Utility.getIntegratedCircuit(2))
```
chochem, line 7293: small problem: the Twilight Forest recipe is broken (probably already before your change). It conflicts with stone->pressure plate which is also on circuit 2

chochem (changes requested):

src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
Cleptomania: It seems to me like the recipe for the pressure plate can be changed to circuit 3 without issue. Is that an acceptable solution?

Cleptomania: Some thought and investigation definitely needs put into making these not conflict, I don't see a super clear answer without digging deeper and probably having to change the other recipes that conflict with it. The recipe for The End one is also conflicting because of the obsidian. Should I try to tackle that as part of this PR or should I make a new issue for these? These were present even before the conversion of the recipes to RA2.

The easy answer is to make them not use stone for TF and not use obsidian for The End, but I'm not sure what suitable replacement blocks might be for those.

chochem: I would leave the pressure plate as its pretty basic and comes from like 20 stone types. the TF thing is super specific and can be on any some other circuit no problem

chochem: the circuits overall look very strange. no reason to go through all numbers. Spoke with GDC that made the recipe. Proposal would be to put _all_ the planet recipes on circuit 17. should be safe and makes more sense.

Cleptomania: I converted them all to circuit 17, none show up in the conflicting recipes now.

Cleptomania:
Important to note before merging that this does now change recipes. The NEI planet recipes have all been changed to use circuit 17, before they each used a different circuit. This was done to prevent conflicting recipes. Nothing else about the recipes has changed.

chochem (approved):
sorry for the sidetrack and thanks!

PR #676, opened 2023-07-16
Fix meta data for EV Lapotron Crystal recipe

chochem:
As most players in IV+ know, you cant just make a pattern for the EV lapo from nei, as then the result wont be recognized but you have to do it manually. This is in fact caused by a small mistake in the recipe.

the recipe is defined without a meta value which defaults to null and actually crafting in the game one gets an item with negative durability (it fixes itself when used but wont be recognized by a normal automation)
image
image
it also shows incorrectly as full in NEI:
image

With this fix instead one gets the correct meta value of 26 corresponding to an empty crystal (durability 1): 
image
image
it also shows correctly in NEI:
image

But not just that, I tested the interaction with NEI and AE2: using the ? to define a pattern from NEI does in fact now work without any adjustments!

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
@@ -931,14 +931,12 @@ public class AssemblerRecipes implements Runnable {
                 GT_ModHandler.getModItem(IronTanks.ID, "goldDiamondUpgrade", 1L, 0),
                 600,
                 120);
-
-        GT_Values.RA.addAssemblerRecipe(
-                new ItemStack[] { CustomItemList.RawLapotronCrystal.get(1L),
-                        GT_OreDictUnificator.get(OrePrefixes.circuit, Materials.Advanced, 2L) },
-                GT_Values.NF,
-                GT_ModHandler.getIC2Item("lapotronCrystal", 1L),
-                600,
-                1024);
+        GT_Values.RA.stdBuilder()
+                .itemInputs(
+                        CustomItemList.RawLapotronCrystal.get(1L),
+                        GT_OreDictUnificator.get(OrePrefixes.circuit, Materials.Advanced, 2L))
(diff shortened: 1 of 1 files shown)


src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
```diff
@@ -931,14 +931,12 @@ public void run() {
...
-                GT_Values.NF,
-                GT_ModHandler.getIC2Item("lapotronCrystal", 1L),
-                600,
-                1024);
+        GT_Values.RA.stdBuilder()
+                .itemInputs(
+                        CustomItemList.RawLapotronCrystal.get(1L),
+                        GT_OreDictUnificator.get(OrePrefixes.circuit, Materials.Advanced, 2L))
+                .itemOutputs(GT_ModHandler.getIC2Item("lapotronCrystal", 1L, 26)).noFluidInputs().noFluidOutputs()
+                .duration(30 * SECONDS).eut(TierEU.RECIPE_EV / 2).addTo(sAssemblerRecipes);
```
boubou19, line 939: TierEU.RECIPE_EV is corresponding to 1920 EU/t, so dividing it by 2 it's 960 EU/t, while the previous recipe had 1024 EU/t. Is it intended? Also, if anything, i'd rather make it run at 1920 EU/t with a 320 seconds duration (same power used in total) What do you think?

chochem: when this last came up I was corrected for using 1024 instead of 960. cant please everyone :)

I really dont care. I will make it 16 seconds at 15/16A then (not 320 seconds lul)

boubou19: oh sorry, did a silly mistake, i meant 320 ticks, so 16s yeah

boubou19 (approved):

PR #680, opened 2023-07-22
More new assembler recipes

chochem:
Added several new assembler recipe where previously there was just a crafting recipe with tools. Mostly covers https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/13563, only some vague or less clear stuff is left. Might be good to close.

Also continued to clean up the assembler recipe file some more (moved recipes away from the soldering alloy method that dont use soldering alloy, added a few more methods for better organisation, unified the ismodloaded checks for the methods, moved the field gens to the rest of the tiered components).

New recipes: Hastelloy-X Block
image

Extreme Engine Intake Casing
image

Steel and Bronze Gear Box
image

Cell Workbench
image

Diamond Gear
image

Needs a small GT PR to add a circuit so that the Steel Gear Box and Solid Steel Machine Casing don't conflict.


Dream-Master (approved):
Looks good so far i can see. We check it indeep in the nev dev version

hallucinogender:
Regular engine intake casings are also missing an assembler recipe at present.
image

PR #686, opened 2023-07-26
Add recipes for crafting input hatches

harrynull:
The tiering:
* Crafting Input Bus (item only): LuV gated by assline
* Crafting Input Slave: UV gated by UV sensor (neutronium?)
* Crafting Input Buffer: UIV/UMV gated by spacetime

Let me know if there are any suggestions / feedback

image
image
image
image
image


harrynull:
Also @TimberDimber told me that they want a upgraded version with more pattern slots. I think depending on the tier of the base block, this can come either free, or with upgrade cards.

src/main/java/com/dreammaster/gthandler/recipes/AssemblingLineRecipes.java
```diff
@@ -1,19 +1,19 @@
 package com.dreammaster.gthandler.recipes;
 
-import static gregtech.api.enums.Mods.GraviSuite;
-import static gregtech.api.util.GT_RecipeBuilder.HOURS;
-import static gregtech.api.util.GT_RecipeBuilder.MINUTES;
-import static gregtech.api.util.GT_RecipeBuilder.SECONDS;
-import static gregtech.api.util.GT_RecipeBuilder.WILDCARD;
+import static gregtech.api.enums.Mods.*;
```
BlueWeabo, line 3: do not have wild imports.

src/main/java/com/dreammaster/gthandler/recipes/AssemblingLineRecipes.java
```diff
@@ -132,6 +132,44 @@ public void run() {
                 .itemOutputs(ItemList.Energy_LapotronicOrb2.get(1L)).eut(TierEU.RECIPE_ZPM).duration(50 * SECONDS)
                 .addTo(AssemblyLine);
 
+        TT_recipeAdder.addResearchableAssemblylineRecipe(
```
BlueWeabo, line 135: and an if (TecTech.isLoaded()) and then call TT_recipeAdder.

src/main/java/com/dreammaster/gthandler/recipes/AssemblingLineRecipes.java
harrynull: yea my IDE loves wildcard imports i guess

Dream-Master:
its idea i have the same problem need to solve it manual now

chochem:
in intellij its under settings->editor->codestyle->java->class count to use import with star: put some high number.

AwonGardino:
May we not gate fluid one THAT far away? Mb just use living solder on that one?

TheEpicGamer274 (approved):
I think the recipe for the fluid one is gated quite nicely

meepmopper (changes requested):
It should be more expensive (idk java at all and have no clue what the recipe is)

Dream-Master:
> It should be more expensive (idk java at all and have no clue what the recipe is)

if you suggested a change you need to name it. You dont need to know java but you need to know GTNH very well

meepmopper:
Sorry, I did not mean to submit that one. I think adding spatially enlarged fluid would be a good change as they are equivalent in tier and makes more sense thematically.

Dream-Master:
@TheEpicGamer274  spatially enlarged fluid instead of space time ?

TheEpicGamer274:
> @TheEpicGamer274 spatially enlarged fluid instead of space time ?

sure, sounds fine

Dream-Master:
@harrynull can you cahnge that?

GDCloudstrike:
there is an active discussion going on in github-discussion right now, please chime in if you have any suggestions

dahgul333:
Ea! lets put liquid STARGATE! insted of spatially enlarged fluid! It will be more fun !(No .and spacetime tier fluid too)
gate it +- on Uhv- start Uev

boubou19:
will probably need a follow up to fine tune the recipes, but overall tiering looks fine

boubou19 (approved):

PR #687, opened 2023-07-27
More optical CPU recipes

GDCloudstrike:
Adds two higher tier optical CPU recipes akin to the optical memory.
These recipes aim to address space assembler speed issues and incentivise higher tier material usage.

First alternate recipe for uiv tier:
image
Second alternate recipe for umv tier:
image


TheEpicGamer274 (approved):
Looks great, ty

PR #696, opened 2023-08-03
fix mistake in brick electrolyze recipe

Pxx500:
In the previous PR about brick electrolyze recipe I made a mistake and put Silicon instead of SiliconDioxide. This should fix oxygen magically turning into silicon


chochem (approved):
The formula is Al2Si4O11 so that 17 of them indeed work out exactly as 5 Al2O3 + 12 SiO2 with chembalance.

Dream-Master (approved):

PR #703, opened 2023-08-17
Remove broken HSLA recipe

chochem:
we dont use rotarycraft, so we dont use HSLA, so this recipe is just broken:
image

2 PRs because the recipe was added twice. The other is https://github.com/GTNewHorizons/GT5-Unofficial/pull/2241


Dream-Master:
@chochem not forget to remove the quest if there is some.

Dream-Master (approved):

chochem:
no uses, so no quest

PR #707, opened 2023-08-19
Mica based plate

Dream-Master:
recipe use a plate mold now


chochem (changes requested):
this doesn't work. the LV and MV forming press only have 2 input slots. (HV is 4)
but we need this in LV.

Dream-Master:
ok this wont work at lv maybe i add it at hv to get a better ratio

chochem:
now there is only the HV one.
If you want both you would also need to make sure they arent conflicting with each other.

Dream-Master:
yea i add it back.

Dream-Master:
i coul add a control circuit for hv one because it has 4 slots

chochem:
they still conflict. The HV one was not the problem, it has the form. but the LV one has nothing to separate it.

chochem:
in fact I just quickly loaded the latest changes here. And in-game only the LV one shows up due to the conflict.

Dream-Master:
i will check ingame if it show a conflict. it will  pick the lv one every time you not setup the control circuit and add a form so i see no conflict here

Dream-Master:
since control circuits not use a extra slot we can add some to lv as well. need to test if that work. so 1 for lv and 2 for hv

Dream-Master:
image
image

chochem (approved):

PR #708, opened 2023-08-19
Mainframe fix

Dream-Master:
IV Mainframe only choose 24 Annelead copper. 
now you can choose between 16 Annealead Copper or 24 Copper cables


chochem:
> now you can choose between 16 Annealead Copper or 24 Copper cables

how do you mean this? you just changed the material to non-annealed, there is no choice.

Dream-Master:
before you can do only the recipe with 24 annealead copper now you should make both recipes.
the recipes got overriden i guess

PR #712, opened 2023-08-20
Recipes for T9 Planet items (EoH catalysts) + DD update

Gofer2x:
Seth, Anubis, Neper, Maahes. The ores are unique to their planets. The red item is a sapling of a tree found on Maahes.
None of the planets have pumpable fluids, so I gave them ones that felt fitting. Not sure about Neper though since its theme is that it's dry (possibly subject to change since it breaks the pattern).
image
Also changes the DD item recipe to use Tartarite Ore instead of Dillithium, since the latter isn't exclusive to DD anymore.
image


Dream-Master:
@Gofer2x its a good idea to add to those planets new fluids to pump. Guess we can rework stuff

Gofer2x:
Oh wait, Tartarite is obtainable from T8 planets lol. I mixed it up with Rubidium (both red). And ye.

GDCloudstrike:
Can you change the recipe voltages for the t9 planets to UIV? 
UMV seems too high because these planets are technically uev tier, so it would be t+2 while the other ones are (mostly) t+1
would also fill the gap between t8 planets (uev recipe) and dd (umv recipe) nicely

Gofer2x:
I just kept them at UMV since the Universium output for them and DD is the same iirc. Yeah, shall do that.

Dream-Master:
@GDCloudstrike

GDCloudstrike (approved):

PR #728, opened 2023-09-20
Add feature to change the icon and title of the minecraft window to GTNH logo and name

Alexdoru:
But implementing it this way, the title and logo will be set from the very beginning of the game, and it allows us to remove this mod from the pack which we only used to do this https://github.com/GTNewHorizons/itlt


src/main/java/com/dreammaster/mixin/mixins/early/MixinMinecraft_PackIcon.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @ModifyArg(
+            method = "startGame",
+            at = @At(
+                    value = "INVOKE",
+                    target = "Lorg/lwjgl/opengl/Display;setTitle(Ljava/lang/String;)V",
+                    remap = false))
+    private String dreamcraft$changeWindowTitle(String original) {
+        this.dreamcraft$loadedGTNHIcon = IconLoader.setCustomIcon("assets/dreamcraft/textures/icon/GTNH_42x42.png");
+        return "GT: New Horizons";
```
Caedis, line 28: Could be nice to have the pack version number here

src/main/java/com/dreammaster/client/util/IconLoader.java
```diff
@@ -0,0 +1,139 @@
+package com.dreammaster.client.util;
+
+import java.awt.*;
```
BlueWeabo, line 3: no wildcards

src/main/java/com/dreammaster/coremod/DownloadProgressDialog.java
```diff
@@ -6,14 +6,16 @@
 import javax.swing.*;
```
BlueWeabo, line 6: i know it already exists as a wildcard, but could clean it up if you want to.

src/main/java/com/dreammaster/coremod/DreamCoreMod.java
```diff
@@ -1,26 +1,29 @@
 package com.dreammaster.coremod;
 
 import java.io.*;
-import java.util.Map;
-import java.util.Properties;
+import java.util.*;
```
BlueWeabo, line 4: don't add more wildcards.

BlueWeabo (commented):
Don't have much else to say really. I have no clue how mixins work. But I am unsure if we want this to be in here or if it should be elsewhere. No clue where else it could be tho. So dunno.

Alexdoru:
> Don't have much else to say really. I have no clue how mixins work. But I am unsure if we want this to be in here or if it should be elsewhere. No clue where else it could be tho. So dunno.

Feature that rebrands minecraft as "GT: New Horizons" -> should be in the mod named "GT: New Horizons"

Alexdoru:
and it allows us to remove the mod "it's the little things"

BlueWeabo:
> > Don't have much else to say really. I have no clue how mixins work. But I am unsure if we want this to be in here or if it should be elsewhere. No clue where else it could be tho. So dunno.
> 
> Feature that rebrands minecraft as "GT: New Horizons" -> should be in the mod named "GT: New Horizons"

I mean i am unsure if we should be adding mixins to a mod where a lot of people edit the code. and unsure if we want to spread out all of our mixins not to centralized mods. That is my only concern and nothing else really.

Alexdoru:
> I mean i am unsure if we should be adding mixins to a mod where a lot of people edit the code. and unsure if we want to spread out all of our mixins not to centralized mods. That is my only concern and nothing else really.

As you said you have no clue how mixins work. This mod already needs mixins to work because it depends on other mods that uses mixins so adding one mixin to it will change nothing.

BlueWeabo:
Alright thanks for the information.

Dream-Master (approved):
mitchej123 — Today at 5:37 PM
looks fine

mitchej123 (approved):
Rebasing master vs merging master would make the commits more clear, since this is also including the confirm exit mixin which I believe was already merged.  But aside from that looks fine.

As a note - we're already doing ASM in this mod, adding in mixins in the core mod isn't a big deal considering unimixins ships with the pack.

mitchej123:
> Unsure if we want to spread out all of our mixins not to centralized mods.

Similar to what Alex said - The mixin(s) or ASM should go in the mod that makes sense.  I agree we don't want them sprawling everywhere, but a mixin to rename the window to GT:NH makes sense in the dreamcraft mod.  Most video/rendering mixins will belong in Angelica going forward, and most performance mixins belong in Hodgepodge.  

As another example - There was a bug rendering fire on water, which really only impacts ThaumicTinker, so I put the mixin there.

Omgise:
@Alexdoru I'm sorry to bother you, but is it possible to separate the window icon replacement feature into a standalone mod?  I really want this functionality 🥺

PR #730, opened 2023-09-23
Cleanup noXXXYYY methods

miozune:


BlueWeabo (approved):
I always forget how many recipes there are here.....

PR #733, opened 2023-09-24
Buff Essentia mirror to maintain consistency with the lore

Alastors:
Effectively, our in-pack lore of the essentia mirror and magic mirror is that the magic mirror is supposed to be one tier higher than the essentia mirror, the magic mirror was originally IV, while the essentia mirror was EV, then the magic mirror was buffed in late 2021:

https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/commit/69d5f23c8afa8e838640e66e7ab91e7e1ccf862f

However, when Piky upgraded the magic mirror from IV to HV back in 2021, he forgot to also buff the essentia mirror from EV to MV respectively, as to fit within our own lore, and this has been bugging me ever since.

No, most reasonable players will not be able to unlock this until HV, since infusion is currently HV, so it's effectively HV anyway, but for consistency's sake, this change needs to happen.

### DIFF
--- a/src/main/java/com/dreammaster/scripts/ScriptThaumcraft.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptThaumcraft.java
@@ -3821,11 +3821,11 @@ public class ScriptThaumcraft implements IScriptLoader {
                 new AspectList().add(Aspect.getAspect("aqua"), 24).add(Aspect.getAspect("iter"), 32)
                         .add(Aspect.getAspect("permutatio"), 16).add(Aspect.getAspect("vitreus"), 8),
                 getModItem(Thaumcraft.ID, "ItemResource", 1, 10, missing),
-                new ItemStack[] { GT_OreDictUnificator.get(OrePrefixes.plate, Materials.Titanium, 1L),
+                new ItemStack[] { GT_OreDictUnificator.get(OrePrefixes.plate, Materials.Aluminium, 1L),
                         GT_OreDictUnificator.get(OrePrefixes.screw, Materials.Thaumium, 1L),
                         GT_OreDictUnificator.get(OrePrefixes.plate, Materials.Steel, 1L),
                         GT_OreDictUnificator.get(OrePrefixes.screw, Materials.Thaumium, 1L),
-                        GT_OreDictUnificator.get(OrePrefixes.plate, Materials.Titanium, 1L),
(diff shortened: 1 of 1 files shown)


mitchej123 (approved):

MatttWasTaken:
Please approve. This would be a very beneficial change. 👍

Dream-Master:
@Alastors Alu is technically doable at lv so why not change this to stainless steel? What material the mirror one tier up use?

Dream-Master:
Other mv Materials are possible too

Alastors:
> Other mv Materials are possible too 

Like? MV doesn't have as many materials as HV from my memory

Dream-Master:
Why not stainless if you think they are no mv materials ? What about magic materials ?

Alastors:
> @Alastors Alu is technically doable at lv so why not change this to stainless steel? What material the mirror one tier up use?

Magic mirrors use enderium which is the same tier as stainless, also boubou has officially stated that within the confines of thaum, stainless is HV, alu is MV and steel is LV. We asked

Alastors:
If you're hard against it, we could use thaumium instead I guess? That's technically MV since you already need to have aluminium to make thaumium

Alastors:
There's not really another appropriate magic material available at MV, and if we are to keep consistency it can't use stainless or it breaks the tier - 1 thing

Alastors:
I will point out, infusion itself is already Hv

Dream-Master:
> If you're hard against it, we could use thaumium instead I guess? That's technically MV since you already need to have aluminium to make thaumium

We hast thaumium in the recipe already

Dream-Master:
> There's not really another appropriate magic material available at MV, and if we are to keep consistency it can't use stainless or it breaks the tier - 1 thing

You dont tell me what the advanced version use

Alastors:
> > There's not really another appropriate magic material available at MV, and if we are to keep consistency it can't use stainless or it breaks the tier - 1 thing
> 
> 
> 
> You dont tell me what the advanced version use

As I mentioned in #meta-dev, there's not an advanced version, the other one is the magic mirror which handles items, this is the essentia mirror that handles essentia.


 
> and since when we have such rule if you not add it ?  

The questbook mentions it, the thaumonomicon mentions it, it's how it was until November 2021, this was just something Piky accidentally missed

Dream-Master (approved):

PR #735, opened 2023-10-01
add SingularityMold

dahgul333:
Singularity mold for creating singularities in Electric Implasion Compressor. firstly for Iron and Fluxed electrum singularities which are used by thousands in endgame( EOH-s and Stargate push). texture a bit crude welcome to change :P
image


Dream-Master:
singularities need to be compressed. Molds not fit here

dahgul333:
> singularities need to be compressed. Molds not fit here
4-D nature of material can do some majic but if it fundamental question
can just rename this to Form

Ethryan:
I would say that this might be better off in a new multi block machine instead of the implosion compressor. Since from what I can remember you sometimes need a lot of blocks to create a singularity. 

Just my opinion

Alastors:
Yeah I agree, new multi

EnderProyects:
Kuba self assigned for a new multiblock for the creation of singularitys almost a year ago xd

dahgul333:
> At least 1 approving review is required by reviewers with write acc

I’ve been hearing about this miracle multiblock for the second year now, but it’s still not there, as soon as it appears, everything can be changed

dahgul333:
> eah I agree, new multi

where can I look at the new multiblock?

EnderProyects:
> > eah I agree, new multi
> 
> where can I look at the new multiblock?

I think someone was waiting for MuTEs to be working fine, and then create the multiblock with that

dahgul333:
a multiblock will not be created soon, but the need is there now

Connor-Colenso:
I appreciate the work, but I also agree and it seems to be the common sentiment that a new multi/mechanics would be appropriate here rather than a mold. So going to close this for now.

dahgul333:
> I appreciate the work, but I also agree and it seems to be the common sentiment that a new multi/mechanics would be appropriate here rather than a mold. So going to close this for now.

eeeey another year without normal singularity craft !

EnderProyects:
The mold could be implemented so we have something to start with, but I also think is better to make a new avaritia multi

EnderProyects:
P.s If someone whants some ideas I come up with this like a year ago ( I tryed to make code for a multi but I failed xd)
image
image
image
image

PR #737, opened 2023-10-07
fix amount of engine intake casings from recipe

Pxx500:
resolves https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14497


Caedis:
Context: Crafted amount was modified in https://github.com/GTNewHorizons/GT5-Unofficial/pull/1082 but assembler amount was not.

Caedis (approved):

PR #738, opened 2023-10-09
Fix fluid canners not working

miozune:
Close https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14676
`mSharedTank` prevents draining from fluid input slot. It looks like we can remove this variable in the future.


Dream-Master (approved):
worked

PR #742, opened 2023-10-16
duration prettier

Quetz4l:
* example: from 128 to 6 * SECONDS + 8 TICKS

### DIFF
--- a/src/main/java/com/dreammaster/scripts/ScriptAdvancedBackpacks.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptAdvancedBackpacks.java
@@ -12,6 +12,7 @@ import static gregtech.api.enums.Mods.SleepingBags;
 import static gregtech.api.util.GT_ModHandler.getModItem;
 import static gregtech.api.util.GT_Recipe.GT_Recipe_Map.sBrewingRecipes;
 import static gregtech.api.util.GT_RecipeBuilder.SECONDS;
+import static gregtech.api.util.GT_RecipeBuilder.TICKS;
 
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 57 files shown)


boubou19:
not sure if we want to change them like that, as you are skipping 8 ticks.

src/main/java/com/dreammaster/scripts/ScriptAdvancedBackpacks.java
```diff
@@ -384,7 +385,7 @@ public void loadRecipes() {
 
         GT_Values.RA.stdBuilder().itemInputs(getModItem(Minecraft.ID, "melon", 1, 0, missing))
                 .fluidInputs(FluidRegistry.getFluidStack("water", 750))
-                .fluidOutputs(FluidRegistry.getFluidStack("melonjuice", 750)).duration(128).eut(4)
+                .fluidOutputs(FluidRegistry.getFluidStack("melonjuice", 750)).duration(6 * SECONDS).eut(4)
```
Dream-Master, line 388: This conversation ist not correct

dahgul333:
some parts are rounded so that the time does not consist of more than 2 parts @boubou19 , @Dream-Master

Dream-Master:
Then round them up not down

Quetz4l:
ah yes, if the duration is more than 100 ticks, the ticks are not written. This is meant to discard useless ticks

GDCloudstrike:
There are a lot of instances where you used x * SECONDS + 10 * TICKS, so why not just keep it as 6 * SECONDS + 8 * TICKS?

Quetz4l:
In the formula with minutes and hours ticks make no sense, they just lengthen the code. you can decide whether it is better to round up or leave it as in the original. tomorrow I will finish the pr

dahgul333:
> There are a lot of instances where you used x * SECONDS + 10 * TICKS, so why not just keep it as 6 * SECONDS + 8 * TICKS? Also, what do you mean by useless ticks? why are they useless?

not all recipes were created with precise parameters, a lot was taken from the head, 8 ticks out of 128 is only 6.2% ,6,2% from a number taken at random, but removes an entire segment

GDCloudstrike:
> 
> not all recipes were created with precise parameters, a lot was taken from the head, 8 ticks out of 128 is only 6.2% ,6,2% from a number taken at random, but removes an entire segment

6.2% is still a significant amount, especially if that is applied to loads of recipes. Scanning through the changes here, it seems that these types of recipes are also comparatively only a small subset of the whole change, so adding those lost ticks back would not significantly impact the code length while keeping accuracy.

dahgul333:
> Then round them up not down

edited 👀

Dream-Master:
did the tick Phrase work? @boubou19 min , sec works.

boubou19 (approved):
i looked 40 recipes randomly with various durations and they all seems fine.

PR #743, opened 2023-10-17
Roasted without tiny/small dusts

dahgul333:
Bulk roasted processing without small/tiny dustes

x4 for ingots

x9 for dusts

https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/11701#issuecomment-1765068496 - self asigned(kek)


OneEyeMaker:
What about dividing by 2 all amounts and duration of ingot recipes? (If it is inline with other recipes)

dahgul333:
> What about dividing by 2 all amounts and duration of ingot recipes? (If it is inline with other recipes)

it is posible for ingots but i am spetialy do this a bit bigger, because - recipes giva a 
base materials and you need it in a bulk , and in order to maintain a closer multiplier relative to crafts with dust, because everything is done with - 4 circuit, I even specifically asked and added .noOptimize() there

OneEyeMaker (approved):

PR #744, opened 2023-10-17
Add stocking hatch recipe

Pelotrio:
Adds a recipe for the stocking input hatch that is gated UHV.
While adding the stocking input hatch in GTNewHorizons/GT5-Unofficial#2322 and some more content on it like GTNewHorizons/GT5-Unofficial#2336 and https://github.com/GTNewHorizons/GT5-Unofficial/pull/2339 we also thought about balancing. While arguably very powerful we don't want it to be gated too far into the late game as it wouldn't make sense with the recipe of the stocking bus. I go more into detail in [this ](https://github.com/GTNewHorizons/GT5-Unofficial/pull/2322#issuecomment-1747539149)comment.

As for the tiering regarding auto pull, the amount of fluids it can pull, and the total amount of fluid I still think it's fine if the full version is unlocked in UHV with downgraded versions potentially getting added later. The hatch tiering in general doesn't matter too much with the amounts you can store in an input hatch in UHV and what you need with the exception being EOH which currently doesn't work with the stocking hatch thus not invalidating the need for the humongous input hatch. Also, functionality like using p2p is not given with the hatch and would require a normal hatch. The rest I discussed in my previously mentioned comment.

In general, the recipe proposed in this PR is not final and should serve as a starting point for a discussion of potential changes but it should also serve as an anchor point so the recipe is not gated way too late into the game like scary comments like [this](https://github.com/GTNewHorizons/GT5-Unofficial/pull/2322#issuecomment-1745488795) suggest. We hope due to the effort we have put into this and the other PRs from https://github.com/Minecraft-TA/ our opinion on this is respected.

The stocking hatch recipe:
image
- 1 Quadruple Input Hatch (UHV)
- 1 ME Dual Interface
- 4 Hyper-Acceleration Card
- 1 Electric Pump (UHV)

Stocking bus recipe for reference:
image

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
@@ -3779,6 +3779,16 @@ public class AssemblerRecipes implements Runnable {
                 .itemOutputs(ItemList.Hatch_Input_Bus_ME.get(1)).duration(15 * SECONDS).eut(TierEU.RECIPE_HV)
                 .addTo(sAssemblerRecipes);
 
+        GT_Values.RA.stdBuilder()
+                .itemInputs(
+                        ItemList.Hatch_Input_Multi_2x2_UHV.get(1L),
+                        GT_ModHandler.getModItem(AE2FluidCraft.ID, "part_fluid_interface", 1, 0),
+                        GT_ModHandler.getModItem(AppliedEnergistics2.ID, "item.ItemMultiMaterial", 4L, 56),
+                        ItemList.Electric_Pump_UHV.get(1L),
+                        GT_Utility.getIntegratedCircuit(1))
+                .itemOutputs(ItemList.Hatch_Input_ME.get(1)).duration(15 * SECONDS).eut(TierEU.RECIPE_UV)
+                .addTo(sAssemblerRecipes);
+
         // Cell Workbench
         GT_Values.RA.stdBuilder()
                 .itemInputs(


Laiff (approved):
Good starting point to discuss balance of this item

miaowwwwww:
i like the pricing. 
me personally would change that 1 quad input hatch uhv to 4, since we have 16 slots in the me hatch, one quad hatch only has 4 slot. 
I don't see much point limiting player's ability to play with fluids after uhv. (unless there is something so broken, plz remind me)
uv would be a bit cheap. 
(btw the ticking speed is too slow for eoh, i would not use it for eoh even if it is usable.)(powergen eohs have better ways than this too)

Pelotrio:
> i like the pricing. me personally would change that 1 quad input hatch uhv to 4, since we have 16 slots in the me hatch, one quad hatch only has 4 slot. I don't see much point limiting player's ability to play with fluids after uhv. (unless there is something so broken, plz remind me) uv would be a bit cheap. (btw the ticking speed is too slow for eoh, i would not use it for eoh even if it is usable.)(powergen eohs have better ways than this too)

That is a good point I also thought about increasing the amount 4 but left it at one for now but this would definitey make sense. I am not familiar with any interaction where the amount would play a significant role that is not achievable with another normal hatch. As for the EOH I think its fine if it just doesn't work with it at least for now.

GDCloudstrike (changes requested):
>The hatch tiering in general doesn't matter too much with the amounts you can store in an input hatch in UHV

I disagree with this point as especially with the updated batch mode, large hatch sizes and thus tiering is more important than ever. Fluid solidifying of low tier materials for example can easily reach hundred thousands of parallel and hatch size is a severely limiting factor in those cases. So no, hatch tiering is not irrelevant.

In your comment on the initial PR you mentioned the stocking hatch not trivializing automation as much as the crafting input buffer, which I also disagree with. The stocking hatch allows for basically the same level of easy automation as the crafting input buffer does while also easily being infinitely expandable. 

This setup is a demonstration of how this would work, the dual interface with the pattern inside is on the orange net and pushes items into the inf chest (or any item storage) and fluids into a dual interface on the crafting subnet via fluid p2p. Then the fluids get stored on fluid drives (or any tank really) and can just be pulled from the network because of auto pull. Same with the items, those can be auto pulled via a storage bus on the inf chest. So in essence you have almost the same expandability as a crafting input buffer does (you can make the subnet arbitrarily large and connect an arbitrary number of machines), with the only downside of needing 1 hatch more on the multi (PA in this case).


For these reasons I would not agree with tiering this any lower than the crafting input buffer. 
Having lower tier versions without auto pull / limit on fluid amount would be fine imo, but the stocking hatch in its current state should not be lower than umv.

JodieRuth:
The auto pull version can be put into UEV, at which point it is already possible to make crafting input buffers (though the stocking hatch doesn't have so many unique features). Versions without this feature may be too late for UHV, and if a version without auto pull is unlocked at UV, it would be great for revamping your factory at this point.

JodieRuth:
The auto pull version can be put into UEV, at which point it is already possible to make crafting input buffers (though the stocking hatch doesn't have so many unique features). Versions without this feature may be too late for UHV, and if a version without auto pull is unlocked at UV, it would be great for revamping your factory at this point.

vovaksenov:
I remember seeing part of this argument somewhere else, but i want to reiterate it here for completeness. Once you reach UHV the cost of most further parts of progression increases drastically compared to the comparatively consistent requirements of LuV-UV. This means it is a time when many would consider redesigning or expanding a significant part of their lines, so having the stocking hatch gated by UHV makes a lot of sense. As for the auto-pull, there definitely is a good case for gating it further, but i hope in case the devs decide to do so, there is a UHV variant without it available as well. I dont think there is currently a significant QoL addition when it comes to I/O between early LuV and late UEV and this is a good opportunity to address this.

miaowwwwww:
> I disagree with this point as especially with the updated batch mode, large hatch sizes and thus tiering is more important than ever.

Are we supposed to suppress machine parallel after uhv? 
if so, we should change the balance of those heavily parallel needed recipes. 
for example, huge parallel needed fluid to ingot, if someone is only blocked from running a recipe "full power" because the hatch size, i think that's on the pack to make a better environment.
u are only forcing people to build more machines for NOTHING.

> The stocking hatch allows for basically the same level of easy automation

crib can do but stoking cannot:1 instant start.2.one click return. 3 parallel material delivery without any wire. 
2 is very important for mobility.  for 3. no wire is much better than the wiring stuff.
if crib only has this infi storage it would not be worth spacetime tier.

(btw why r u using this much set up, just two dual interface facing each other and let drivers do the work, u don't need p2p u don't need infi chest. u don't need a third network)

GDCloudstrike:
> The auto pull version can be put into UEV, at which point it is already possible to make crafting input buffers (though the stocking hatch doesn't have so many unique features). Versions without this feature may be too late for UHV, and if a version without auto pull is unlocked at UV, it would be great for revamping your factory at this point.

It is not possible to make crafting input buffers at uev as they require spacetime, that means they are locked to late uiv / early umv.


Versions without auto pull being unlocked earlier makes sense and I would agree with what sampsa said on that regard:

> A single-slot, non-autopull stocking hatch tiered at EV-IV would be fine, with maybe 4-slot in IV-LuV.

Adding onto that, a 16 slot version in ZPM-UHV would also make a lot of sense to make I/O easier but not auto-pull levels of easy. 


>for example, huge parallel needed fluid to ingot, if someone is only blocked from running a recipe "full power" because the hatch size, i think that's on the pack to make a better environment.

That is precisely why larger hatches exist. If your machines are capping out your uhv hatches you have a reason to upgrade to uev ones. If you suddenly get an infinitely large hatch at an early tier then that part of progression becomes completely irrelevant for the rest of your playthrough.

>crib can do but stoking cannot:1 instant start.2.one click return. 3 parallel material delivery without any wire.
2 is very important for mobility. for 3. no wire is much better than the wiring stuff.
if crib only has this infi storage it would not be worth spacetime tier.

No.1 is not really a make or break kind of deal, its helpful sometimes, mainly on smaller crafts, but does not really make any meaningful difference on larger ones when your machines are running almost constantly anyways. 

No.2 is a similar matter as that only is relevant when a recipe is encoded wrong or there is a recipe conflict, but that is usually a one time fix and that's it. One exception might be the dtpf but stocking hatches allow for input from other hatches as well so some catalyst remaining is not relevant in this case.

Out of these three differences I'd argue that no.3 is the most impactful one, that you don't have to wire crafting input buffers. This of course makes their setup easier by a decent margin, but as you pointed out, my setup can be simplified even further meaning that only really the wiring matters in a comparison between the two. That begs the question, is having to place some cables and wireless connectors really a valid argument for downtiering the hatch be 2-3 whole tiers? I don't believe that to be the case.

vovaksenov:
As long as there are options without autopull gated at UHV or before, i dont mind gating the autopull one as far as necessary as follows from the reasons provided by GDCloudstrike.

miaowwwwww:
as GDC said:"I'd agree with a single slot non autopull as early as EV/IV tbh" "and then 4 slot at LuV/ZPM, 16 slot UV/UHV" 
auto pull about UEV?
I would happily see ev / luv/ uv, variants and an auto pull upgrade card maybe.

vovaksenov:
After a discussion in #beta-testing the following tiering was deemed reasonable:
- IV gated 1 slot hatch / 128.000L capacity
- LuV (Fusion MK1) gated 4 slot hatch / 512.000L capacity per slot
- UHV gated 16 slot hatch / 4.096.000L capacity per slot
- UIV transmetal gated 16 slot autopull hatch / 16.384.000L capacity per slot
- UXV 16 slot humongous autopull hatch / 2.1GL capacity per slot
Discussion reference:
https://discord.com/channels/181078474394566657/522098956491030558/1164212941336350760

dahgul333:
if you want
 balance the throughput, do this in machines and not in every hatch

RaitGamerGR:
Why most usefull me hatch ever need to be locked after chemistry, autocraft, crafting lines? Stocking hatch need to be locked on ev-iv. On tier, when you star building your base with automated processes. Do you realy want(like?) to rebuild your base every tier? Do you want to do automation using ineffective and difficult methods? If you ansvered "yes" to the most question, ask you, why stocking bus locked only on HV? Where it the different betwen fluids and items? There is only 1 different. Split possibility. You can easy split items for most recipes, but you cant easy split fluid without container, because most recipes uses not minimum part of fluid (1mb). But most usable item recipes uses minimum amount of item (1 thing). This is the reason why useless crafting input buffer locked on iuv, but crafting input bus on luv. For split, use and move items on luv wave stocking bus, linked bus, p2p interfaces. For fluds we have 1 effective(fluid p2p) and some ineffective methods(tanks(cells)/packets, dual p2p).

About tiered hatches, it will be useless at same tier, because with 5 secs timeout it wil be slower fluid provider than input hatch+export bus, or just use certus tanks/any other buffer to store and provide all fluids. Only uhv hatch will be faster then unloading certs tanks into regular hatch.

> * IV gated 1 slot hatch / 128.000L capacity
> * LuV (Fusion MK1) gated 4 slot hatch / 512.000L capacity per slot
> * UHV gated 16 slot hatch / 4.096.000L capacity per slot
> * UIV transmetal gated 16 slot autopull hatch / 16.384.000L capacity per slot
> * UXV 16 slot humongous autopull hatch / 2.1GL capacity per slot

vovaksenov:
> About tiered hatches, it will be useless at same tier, because with 5 secs timeout it wil be slower fluid provider than input hatch+export bus, or just use certus tanks/any other buffer to store and provide all fluids. Only uhv hatch will be faster then unloading certs tanks into regular hatch.

I already mentioned it to you on discord but to make sure everyone sees the whole picture, the buffer is refreshed immediately upon recipe check. This means that the fluid input is not limited by the 5 seconds cooldown unless the machine is idle. This means that the buffer size is more in order to allow for higher parallel and related stuff rather than to directly limit io speed.

Pelotrio:
@GDCloudstrike I want to focus mainly on what you said because you were on of the few people who actually responded to what I said originally and have actually tested something. I have tried listing and summarizing all the points you made so I can respond to them individually

**1. Hatch size matters because of the new batch mode and very large parallels on low tier materials e.g fluid solidifying
2. Example setup for and infinitely expandable setup with fluids and items
3. Stocking hatch trivializes automation as much as CRIB
4. Crafting Input buffer is late UIV to UMV**


## Hatch size matters because of the new batch mode and very large parallels on low tier materials e.g fluid solidifying
This is partly true but not quite. I have tested processing arrays with the full amount of machines in tiers UHV-UMV with fluid solidifying tin ingots and none of them where able to completely empty one hatch of equal tier with batch mode enabled. And this is without considering that you can use multiple hatches or sometimes be able to craft higher tier hatches in lower tiers. Additionally solidifying tin is one of the lowest tier materials there is and downtiering is only an option in UEV+. (all machines UEV+ were tested with downtiering enabled as well). I would argue that affording a hatch in the voltage tier you are is a non issue so I have not included the cost of hatches. To be fair I cant really say much about tiers that are UIV+ because I have not reached that voltage tier yet. I can still understand that it could be difficult to keep the input hatches always full and that this could be a concern.

##  Example setup for and infinitely expandable setup with fluids and items
What you build is quite a nice setup that is infinitely expandable and supports items and fluids. 
Here is one that does same but is available in EV:

This one uses the same mechanic for stocking items but distributes the fluids using a buffer tank a pump and a fluid p2p. The only downside of this setup is that the amount of fluid in the pattern has to be divisible by amount of machines. 

Here is another setup for good measure:

This setup uses dual interface p2ps and advanced blocking card to evenly distribute the fluids and items and only inserting one set at a time and is also infinitely expandable. 
You could argue that both setups probably suffer with very fast machines but that can be fixed with batch sizes. 


Here is my setup with the same functionality using stocking hatches.
java_WETCFK6Jhn
It works very similar to yours but if you compare it to the other two setups it only uses a few less components for a pretty similar result. 

Now lets check out the same functionality with the CRIB:
 

java_0NdWSEkS6I
That is literally all you need which brings me to my next point

## Stocking hatch trivializes automation as much as CRIB
That is not really true if you actually compare the two. 

CRIB has/is
- infinitely expandable using slaves that are pretty free to craft once you actually get one
- Instant recipe start
- Doesn't need wire connection
- Needs one casing slot
- Has complete pattern isolation

Stocking hatch has/is
- Infinitely expandable but not so easy to craft when you first unlock it
- Needs to wait for recipe check of the connected machine
- Needs wire connection
- Needs two casing slots most of the time because most of the time a stocking bus is required as well
- Can potentially conflict in the same subnet depending on the machine

Mainly the casing amount and wire connection here matters as it limits how many different setups you can connect due to the need for multiple hatches and busses for different circuits and other non consumables. With the CRIB you literally have to add one block per machine.

If you compare the setup you can also see that the infrastructure needed to use a setup like this shrinks dramatically and it does it very gradually. Which for me makes it obvious for me that this should be available somewhere in between the p2p/dual p2p (EV) setup and the CRIB (late UIV - UMV??????).

## Crafting Input buffer is late UIV to UMV
That is simply not true at all it requires spacetime which is available with a DTPF with hypogen coils which is late UEV. I am happening to be in that state of the game at the moment and I will be able to craft one very soon. You could argue you can make UIV components with the tesseracts you have to craft before but it is definitely possible to achieve without crafting a single UIV component except circuits. I don't know if that was different when you reached that stage or if you misremember but its definitely not the case.

## Thoughts on tiered stocking hatches in general
After thinking all of this through I don't think its a good idea to make stocking hatches available for lower tiers at least not at a the level of the stocking bus. My reasons for that are

- I find the idea of having about 5 variants of the same block annoying even if it would have some kind of upgrade system
- None of the variants have been programmed yet and would take a bit to implement (this is more personal preference and not really a valid point to not do it)
- The metrics of the tiering system doesn't really block anything except autopull
- Low tier stocking hatch in lower tiers would make Chemlines substantially easier
- The stocking bus is literally gated EV

I'll try to expand on the last three points here a but more. 
If you for example get a one slot stocking hatch in IV going forward you can literally build every chemline using a central storage with item and fluid storage, build all required machines and stock every item and fluid in the correct machine. This way you dont have to worry about anything as all fluids and items stay in the central system unless they are actively used. if different machine speed is an issue you can also just have a normal output bus/hatch and place the storage bus there and it auto limits the machine as well due to the new void modes. The amount of fluids the hatch can pull is also a non issue as you can literally just build multiple hatches what you would do anyway except when using quad hatches. The tiering between 4 and 16 fluid slots also does basically nothing as there are almost no recipes that use more than 4 fluids unless you do hatch sharing which you also could just avoid. As for the amount the hatches can pull that is not really relevant until late into the game in my experience. (I literally used MV-EV hatches up until like ZPM as no machine can process that much at once). In my opinion this would only be relevant at the very late game which seems to be the perspective you have on things.

As for the last point the stocking hatch is literally not gated at all. You unlock it the moment you can use AE2 and its arguably even more powerful than the fluid variant as items are used very much in the early game. if the stocking hatch would be gated UIV or UMV there would be 7-8 voltage tiers between the item variant and the fluid variant which would be totally insane.


## My proposal
As we have found out the CRIB is way more powerful than the stocking hatch and it is only gated late UEV so it would make sense that the stocking hatch is gated before that. My initial recipe is gated early UHV so my suggestion would be:

- Gate the stocking hatch one tier before CRIB so late UHV

Pelotrio:
@GDCloudstrike I want to focus mainly on what you said because you were one of the few people who responded to what I said originally and have done some testing. I have tried listing and summarizing all the points you made so I can respond to them individually

**1. Hatch size matters because of the new batch mode and very large parallels on low-tier materials e.g fluid solidifying
2. Example setup for an infinitely expandable setup with fluids and items
3. Stocking hatch trivializes automation as much as CRIB
4. Crafting Input buffer is late UIV to UMV**


## Hatch size matters because of the new batch mode and very large parallels on low-tier materials e.g fluid solidifying
This is partly true but not quite. I have tested processing arrays with the full amount of machines in tiers UHV-UMV with fluid solidifying tin ingots and none of them were able to completely empty one hatch of equal tier with batch mode enabled. This is without considering that you can use multiple hatches or sometimes be able to craft higher-tier hatches in lower tiers. Additionally solidifying tin is one of the lowest-tier materials there is and down-tiering is only an option in UEV+. (all machines UEV+ were tested with down-tiering enabled as well). I would argue that affording a hatch in the voltage tier you are in is a nonissue so I have not included the cost of hatches. To be fair I can't say much about tiers that are UIV+ because I have not reached that voltage tier yet. I can still understand that it could be difficult to keep the input hatches always full and that this could be a concern.

##  Example setup for an infinitely expandable setup with fluids and items
What you build is quite a nice setup that is infinitely expandable and supports items and fluids. 
Here is one that does the same but is available in EV:

This one uses the same mechanic for stocking items but distributes the fluids using a buffer tank a pump and a fluid p2p. The only downside of this setup is that the amount of fluid in the pattern has to be divisible by the number of machines. 

Here is another setup for good measure:

This setup uses dual interface p2ps and advanced blocking card to evenly distribute the fluids and items and only insert one set at a time and is also infinitely expandable. 
You could argue that both setups probably suffer with very fast machines but that can be fixed with batch sizes. 


Here is my setup with the same functionality using stocking hatches.
java_WETCFK6Jhn
It works very similar to yours but if you compare it to the other two setups it only uses a few fewer components for a pretty similar result. 

Now let's check out the same functionality with the CRIB:
 

java_0NdWSEkS6I
That is literally all you need which brings me to my next point

## Stocking hatch trivializes automation as much as CRIB
That is not true if you compare the two. 

CRIB has/is
- infinitely expandable using slaves that are pretty free to craft once you get one
- Instant recipe start
- Doesn't need a wire connection
- Needs one casing slot
- Has complete pattern isolation

The stocking hatch has/is
- Infinitely expandable but not so easy to craft when you first unlock it
- Needs to wait for recipe check of the connected machine
- Needs wire connection
- Needs two casing slots most of the time because most of the time a stocking bus is required as well
- Can potentially conflict in the same subnet depending on the machine

Mainly the casing amount and wire connection here matters as it limits how many different setups you can connect due to the need for multiple hatches and busses for different circuits and other nonconsumables. With the CRIB you literally have to add one block per machine.

If you compare the setup you can also see that the infrastructure needed to use a setup like this shrinks dramatically and it does it very gradually. This for me makes it obvious that this should be available somewhere in between the p2p/dual p2p (EV) setup and the CRIB (late UIV - UMV??????).

## Crafting Input buffer is late UIV to UMV
That is simply not true at all it requires spacetime which is available with a DTPF with hypogen coils which is late UEV. I am happening to be in that state of the game at the moment and I will be able to craft one very soon. You could argue you can make UIV components with the tesseracts you have to craft before but it is definitely possible to achieve without crafting a single UIV component except circuits. I don't know if that was different when you reached that stage or if you misremember but it's definitely not the case.

## Thoughts on tiered stocking hatches in general
After thinking all of this through I don't think it's a good idea to make stocking hatches available for lower tiers at least not at a level of the stocking bus. My reasons for that are

- I find the idea of having about 5 variants of the same block annoying even if it would have some kind of upgrade system
- The metrics of the tiering system don't block anything except auto-pull
- Low-tier stocking hatch in lower tiers would make Chemlines substantially easier
- The stocking bus is literally gated EV

I'll try to expand on the last three points here a bit more. 
If you for example get a one-slot stocking hatch in IV going forward you can literally build every chemline using a central storage with item and fluid storage, build all required machines, and stock every item and fluid in the correct machine. This way you don't have to worry about anything as all fluids and items stay in the central system unless they are actively used. If different machine speed is an issue you can also just have a normal output bus/hatch and place the storage bus there and it auto-limits the machine as well due to the new void modes. The amount of fluids the hatch can pull is also a nonissue as you can literally just build multiple hatches which you would do anyway except when using quad hatches. The tiering between 4 and 16 fluid slots also does basically nothing as there are almost no recipes that use more than 4 fluids unless you do hatch sharing which you also could just avoid. As for the amount the hatches can pull that is not really relevant until late into the game in my experience. (I used MV-EV hatches up until ZPM as no machine can process that much at once). In my opinion, this would only be relevant at the very late game which seems to be the perspective you have on things.

As for the last point, the stocking bus is not gated at all. You unlock it the moment you can use AE2 and it's arguably even more powerful than the fluid variant as items are used very much in the early game. It also doesn't have any kind of limitation to pull amount (unfair because super busses exist), has auto pull enabled by default, and has 16 slots unlocked.
if the stocking hatch would be gated UIV or UMV there would be 7-8 voltage tiers between the item variant and the fluid variant which would be totally insane.


## My proposal
As we have found out the CRIB is way more powerful than the stocking hatch and it is only gated late UEV so it would make sense that the stocking hatch is gated before that. My initial recipe is gated early UHV so my suggestion would be:

#### Gate the stocking hatch one tier before CRIB so late UHV.
My suggestion would be to tweak the recipe to instead use 4 quad neutronium fluid pipes and additionally add radox polymer. This would force you to build Xenoxene and radox line if you want to use it and push it back after the tier 8 rocket as well. The UEV gate with Amunra is pretty rough nowadays requiring Radox, Kevlar,  the start of Monazite line, and about two stacks of UHV field generator. I have experienced this myself and already heard of it from multiple people who felt the same. It makes sense that you would do a major infra upgrade/rebuild then and if you are willing to build radox then I don't see a reason you shouldn't have it by then. Just as a reminder the majority of players have never reached late UHV at all so it is already pretty late.

#### Make the amount the stocking hatch can pull relative to the voltage tier of the machine.
If the amount of fluid the hatch can pull is still seen as an issue I would offer a different solution. Instead of having different hatches that have different amount of pull amounts instead the voltage tier of the machines determines how much it can consume at once when the machine starts. So a machine running at IV power could pull 256,000 liters a machine running at LuV power 512,000 liters and so on you get the idea. I already checked and this would be possible to implement so that every time the machine starts it can only consume as much fluid as an equivalent hatch of that voltage tier. I don't know if that solution is perfect and if there are exceptions to this but it sounds pretty good to me. I would further say the space of the hatch only matters in very high tiers where CRIB is available anyway which doesn't have any kind of limitation either.

If tiering the hatch is still intended despite my points against it I would say the only thing that would really make a difference is a one-slot non auto-pull stocking hatch that is gated IV-LuV (maybe past assline).

## tldr for anyone too lazy to read
- Hatch sizes matter only partly in later stages of the game and could be solved by making pull amount relative to machine voltage tier
- Similar setups can be achieved way earlier in the game using more components
- CRIB definitely keeps the title of trivializing the most
- CRIB itself is only late UEV and not late UIV or UMV

Steelux8:
> After a discussion in #beta-testing the following tiering was deemed reasonable:
> 
>     * IV gated 1 slot hatch / 128.000L capacity
> 
>     * LuV (Fusion MK1) gated 4 slot hatch / 512.000L capacity per slot
> 
>     * UHV gated 16 slot hatch / 4.096.000L capacity per slot

I suggest an early LuV, after-assline for one of these tiers, and definitely not an Mk1 fusion tier. Fusion is already supremely powerful as a bonus to being required for progression, whereas the assline gives the player very little of an upgrade besides just allowing them to progress into LuV.

vovaksenov:
I find the "Make the amount the stocking hatch can pull relative to the voltage tier of the machine" idea rather unwelcome if you are only leaving 1 tier, as there were definitely cases when i had to use hatches several tiers larger than the machine voltage (and im sure this only gets worse the further you go) and this would make the supposedly late game content potentially worse than the normal hatches by design. Better lock it further but without throttling it. And since it likely is getting locked further, at least one tier should exist before it. I admit that i find it difficult to correctly predict the full implication on the balance, and i agree, that yesterdays list started with a rather early IV option that is not fully appropriate, but i feel like by early UV players should at least have the "inferior" 1 slot stocking option (no autopull; and whether this one is throttled or not and how is not as important for me). Many of the chem lines are completed by this point, so the trivialization aspect should be mitigated at least. More experienced members can decide on what the precise gate for the inferior option(s) should be.

Pelotrio:
> I find the "Make the amount the stocking hatch can pull relative to the voltage tier of the machine" idea rather unwelcome if you are only leaving 1 tier, as there were definitely cases when i had to use hatches several tiers larger than the machine voltage (and im almost sure this only gets worse the further you go) and this would make the supposedly late game content potentially worse than the normal hatches by design. Better lock it further but without throttling it. And since it likely is getting locked further, at least one tier should exist before it. I admit that i find it difficult to correctly predict the full implication on the balance, and i agree, that yesterdays list started with a rather early IV option that is not fully appropriate, but i feel like by early UV players should at least have the "inferior" 1 slot stocking option (no autopull; and whether this one is throttled or not and how is not as important for me). Many of the chem lines are completed by this point, so the trivialization aspect should be mitigated at least. More experienced members can decide on what the precise gate for the inferior option(s) should be.

Me on the other hand never had this problem and even if this becomes an issue you can still use normal hatches. That would be one thing that makes it inferior in some cases and forces you to use hatches/be speed limited. Also there are not that many tiers where you can actually make hatches that are several tiers larger. Additionally you can always just add more hatches

Pelotrio:
Due to the balance changes and the addition of different types of stocking buses/hatches I will close this pull request. Recipes for all types have been added here https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/748

PR #753, opened 2023-11-10
Add lapotronic energy orb cluster space assembler recipe

GDCloudstrike:
Adds a space assembler recipe for lapotronic energy orb clusters to keep it consistent with the other energy orbs and to make faster than 1 tick production of these possible.

image

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/SpaceAssemblerRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/SpaceAssemblerRecipes.java
@@ -108,6 +108,26 @@ public class SpaceAssemblerRecipes implements Runnable {
                 }
             }
             if (GTPlusPlus.isModLoaded()) {
+
+                // Alternate Energy Orb Cluster Recipe
+                IG_RecipeAdder.addSpaceAssemblerRecipe(
+                        new ItemStack[] { ItemList.Circuit_Board_Multifiberglass.get(1L),
+                                GT_OreDictUnificator.get(OrePrefixes.foil, Materials.NaquadahAlloy, 64L),
+                                GT_OreDictUnificator.get(OrePrefixes.circuit.get(Materials.Master), 4L),
+                                ItemList.Circuit_Parts_Crystal_Chip_Master.get(64L),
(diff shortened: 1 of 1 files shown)


miaowwwwww:
good stuff

Pelotrio:
I am definitely a fan of adding a recipe for this in the space assembler but I dont think its a good idea to just use the assembly line recipe for it. This is not consistent with the other recipes for energy cluster which all are a bit cheaper but require more high level material. Due to the usages of the energy orb cluster in higher tier recipes I would recommend to change the recipe to be consistent with the other cluster types.

GDCloudstrike:
> I am definitely a fan of adding a recipe for this in the space assembler but I dont think its a good idea to just use the assembly line recipe for it. This is not consistent with the other recipes for energy cluster which all are a bit cheaper but require more high level material. Due to the usages of the energy orb cluster in higher tier recipes I would recommend to change the recipe to be consistent with the other cluster types.

the reason I decided to not make this recipe cheaper is because this already _is_ the cheaper recipe that uses less higher tier materials. 
If you look at the recipe structure, the cheaper higher tier clusters are based on this recipe, so they are inherently consistent because their basic structure aligns with this recipe.

BlueWeabo (approved):
Thanks for your work

Dream-Master:
@Pelotrio feel free to add a new recipe or change the existing ones.

Pelotrio:
> > I am definitely a fan of adding a recipe for this in the space assembler but I dont think its a good idea to just use the assembly line recipe for it. This is not consistent with the other recipes for energy cluster which all are a bit cheaper but require more high level material. Due to the usages of the energy orb cluster in higher tier recipes I would recommend to change the recipe to be consistent with the other cluster types.
> 
> the reason I decided to not make this recipe cheaper is because this already _is_ the cheaper recipe that uses less higher tier materials. 
> If you look at the recipe structure, the cheaper higher tier clusters are based on this recipe, so they are inherently consistent because their basic structure aligns with this recipe.

These are valid points but I am still not sure if it should be completely identical. Also does the assembler oc the recipe? What about making it a recipe with more Inputs and more outpus eg. 4

Dream-Master:
i am not sure if the item is stackable?

PR #759, opened 2023-11-25
Remove duplicate circuit recipes from code

chochem:
All these recipes are already auto-removed (as collisions) in all the recent stables. They are no longer needed since circuit components got GT oredicts. So it is best to remove them directly from the code.

Also cleaned up the order and comments slightly.

I checked the recipes with this PR in a full game and they all look fine.

This is related to https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14935 but not the main issue there.


Dream-Master (approved):

Dream-Master:
works and tested

PR #763, opened 2023-11-30
reduce cost of coal ball to what it should be

Glease:
one flint dust is 9 flint item, not 1 flint item. using dust also does not match the item texture.

fix https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/11445


bombcar:
Isn't 1 flint dust = 2 flint in the macerator? Not that it really matters ...

Dream-Master:
This recipe get overridden by gt code. Need a fix.

boubou19:
Do you know where the recipe is in GT5U? So we can remove it.

chochem:
will do

chochem:
resolved by https://github.com/GTNewHorizons/GT5-Unofficial/pull/2534

PR #771, opened 2023-12-06
change essentia cells index for ThE

MCTBL:
https://github.com/GTNewHorizons/ThaumicEnergistics/pull/56


MCTBL:
Change the recipes for essentia cells, because of id's change.

PR #772, opened 2023-12-07
remove Ae2 Rockcrusher recipes

Dream-Master:
udate deps
Fix https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15010
(maybe disable mod support in AE2 cfg work too?)


Dream-Master:
ok seems only be the crusher affected
image

PR #775, opened 2023-12-10
Recipe remove NPE

ghostflyby:
There is such output stack whose getItem returns null

### DIFF
--- a/src/main/java/com/dreammaster/recipes/RecipeRemover.java
+++ b/src/main/java/com/dreammaster/recipes/RecipeRemover.java
@@ -53,7 +53,13 @@ public class RecipeRemover {
         int i = tList.size();
         tList.removeIf(r -> {
             ItemStack rCopy = r.getRecipeOutput();
-            if (rCopy == null) return false; // ????????????????????
+            if (rCopy == null) {
+                return false;
+            }
+            if (rCopy.getItem() == null) {
+                System.out.println("Do you know what youare doing?");
+                return false;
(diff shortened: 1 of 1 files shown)


src/main/java/com/dreammaster/recipes/RecipeRemover.java
```diff
@@ -53,7 +53,13 @@ private static void stopBuffering() {
         int i = tList.size();
         tList.removeIf(r -> {
             ItemStack rCopy = r.getRecipeOutput();
-            if (rCopy == null) return false; // ????????????????????
+            if (rCopy == null) {
+                return false;
+            }
+            if (rCopy.getItem() == null) {
+                System.out.println("Someone is adding recipes with null items!");
```
eigenraven, line 60: Use the logger like below, not system.out:
```suggestion
                MainRegistry.Logger.warn("Someone is adding recipes with null items!");
```

boubou19: might also be a good idea to throw a stacktrace in the log to see where the recipe is

ghostflyby: Should I log it in warn (or other log level) or directly `Thread.dumpStack()`?

eigenraven: Log it in warn

boubou19 (approved):

PR #777, opened 2023-12-11
Add advanced radiation proof plate space assembler recipe

GDCloudstrike:
Adds a one step space assembler recipe for advanced radiation proof plates to shift it away from asslines/aals.

(there's also some molten lead in there)
image


Pelotrio (approved):
Implemented as discussed

Dream-Master (approved):

PR #779, opened 2023-12-13
Essentia Condenser Nuke #2

Alastors:
As promised, the second something that could feasibly replace the condenser gets introduced, I'd nuke the condenser, here that is.

I'm going to leave the condenser itself in-tact as a block in-case someone actually wants to fix it in the future (this just removes the recipe replacements from coremod, base recipes were handled as well)

Sister PR to:
https://github.com/GTNewHorizons/MagicBees/pull/34 
and
https://github.com/GTNewHorizons/Gadomancy/pull/26

Fixes:
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10522


Dream-Master:
@Alastors since Element of darkness have no use we can use it for the new crafting recipe as well.

Alastors:
> @Alastors since Element of darkness have no use we can use it for the new crafting recipe as well.

I genuinely don’t think we should do that, IMO the essentia bee is already going to be a little weak, likewise it wouldn’t really make sense since it’s not really a 1:1 replacement

Dream-Master:
@Alastors i set it draft

Dream-Master:
Not needed anymore

PR #788, opened 2024-01-05
Various recipe fixes and tweaks

AbdielKavash:
I went through github issues and my own notes and collected a set of various minor recipe issues and requests, and implemented all I could find. None of these should have a significant balance impact.

* [Nuclear Control] Fixed Advanced Panel Extender recipe. Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15241.
* [Thaumic Bases] Added recycling recipes for Quicksilver Block and Bricks. Both smelt into 9 Quicksilver, or fluid extract into 9000 Mercury. Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14913.
* [ProjectRed] Added recipes for missing components: [Item Crafting Extension Chip](https://i.imgur.com/TA1UonC.png) ([circuit assembler recipe](https://i.imgur.com/EnXDRUX.png)) and [Network Latency Pipe](https://i.imgur.com/Xcdixx9.png). Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15262 and https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14915. I decided not to add recipes for Pressure tubes, since they basically overlap in function with GT item pipes. If somebody wants them because they look cool, they can be added.
* [Tinker's Mechworks] Added an [assembler recipe](https://i.imgur.com/aw0Q9ba.png) to directly craft a full Spool of Wire. To avoid conflicts, the original recipe for an empty spool now requires circuit 1. Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/12053 in a satisfactory way.
* [Logistics Pipes] The recipe for a Blank Module is no longer just a conversion from Project Red Null-Logic Routing Chip. Basic ingredients remain the same, but the Blank Module can now be [crafted directly](https://i.imgur.com/TIPAUXM.png). However, since it doesn't make sense for PR components to be in the recipe anymore, the circuit plate has been replaced by iron plates instead. Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14535 (which was actually a recipe conflict issue).
* [Steve's Carts] The Blank Upgrade now takes steel plates, instead of reinforced plates. This makes basic upgrades for the Cart Assembler available in early LV, without requiring an implosion compressor. Note that many advanced upgrades still require reinforced metal on top of the blank upgrade, this has not been changed. Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14655.
* [Hardcore Ender Expansion] Added recipes for many decorative blocks from the mod. Most blocks have both a crafting recipe, and a (usually more efficient) assembler recipe. Variants of Ravaged Brick and Persegrit are chiselable. [Full list of added recipes](https://i.imgur.com/zvp1bFl.png). Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15093. (*I don't know if this would be better as a separate PR in HEE itself, but that repo seems to be very dead.*)
* [Chisel] Added mixer recipes for all colors of hempcrete powders. Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14886.
* [Chisel] Added more efficient bulk crafting recipes in the assembler for the remaining blocks that still didn't have them: Anti Block, Carpet (includes cutting machine recipes), Cubits, Fantasy Block, Neonite, GloTek, Rebellious Block, Structural Concrete, Purpur Block, Svelstone, Woolen Clay. All new recipes follow the same pattern as existing ones, where the amount of "cheap" input materials (stone, iron, ...) is cut about in half.

I did my best to make sure that none of the added/changed recipes have any recipe conflict in the full pack.


chochem:
why are they all added in scripts when its mostly normal gt recipes?

src/main/java/com/dreammaster/scripts/ScriptLogisticPipes.java
```diff
@@ -1262,6 +1256,114 @@ private void recipes1() {
...
+                        ItemList.Circuit_Basic.get(1L),
+                        GT_OreDictUnificator.get(OrePrefixes.wireGt01, Materials.Gold, 4L),
+                        GT_Utility.getIntegratedCircuit(18))
+                .itemOutputs(getModItem(LogisticsPipes.ID, "item.itemModule", 4, 0, missing))
+                .fluidInputs(FluidRegistry.getFluidStack("molten.lead", 144)).duration(30 * SECONDS)
+                .eut(TierEU.RECIPE_LV).addTo(assemblerRecipes);
+        GT_Values.RA.stdBuilder()
+                .itemInputs(
+                        GT_OreDictUnificator.get(OrePrefixes.plate, Materials.Iron, 4),
+                        ItemList.Circuit_Microprocessor.get(1L),
```
chochem, line 1280: is this on purpose? These itemlist circuits should only be used if you want to forbid unification (multiple occurences, not just here)

AbdielKavash:
I tried following the adjacent code for similar recipes as closely as I could, both for placement and for methods used. Is there some documentation that I could read for the "proper" way of doing this? So far I'm trying to learn by following existing code.

chochem:
not really. Scripts folder are mostly autoconverted things though, so often not ideal, unless they got a good cleanup.
If you look at the itemlist you use, you can see it specificly has the no-unification tag. So if you want to allow all circuits of the tier  (which is almost always the case) you cant use that but should use the gt oredictunificator for example.
When it comes to where recipes go our current idea is to put gt machine recipes in the folder for that machine. Required mods are then just checked for a small section of recipes or even individually directly now with the new option miozune added to RA2.

AbdielKavash:
What would you recommend then? Should I keep the new recipes together with the old ones, for consistency? Or move them to the respective machine-specific files? Or even move *everything* (both newly added recipes and the old ones) to where they should be now?

chochem:
I would move just move them.
Moving them all would also be nice but is a lot of work and might also include some more cleanup. I might get back to that at some point.

chochem (changes requested):
well the moving is one thing. And since the file is not cleaned up anyway, its not critical.

But you need to fix the circuits.

AbdielKavash:
I have cleaned up the circuits and other OreDict nonsense in Project Red and Logistics Pipes. I also caught two minor errors, the time of one recipe was obviously wrong; and two recipes in Logistics Pipes were missing a programmed circuit. I have manually verified the changes side-by-side to make sure nothing went missing.

I agree that as a whole the coremod recipes need a hug amount of cleanup and reorganization. But that should be done separately in a different PR.

chochem (approved):
ah you cleaned up the entire 2 files now and not just your recipes. thats awesome!

I dont see any mistakes or missing stuff right now but of course its a lot.

PR #794, opened 2024-01-17
Allow customizing  Tinker's Construct manuals in dreamcraft

YannickMG:
This PR is unfinished  pending the inclusion the required lang file keys to allow for translation.

Actual customization might happen in a different PR, or in this one.  More is done but it is not yet cleaned up for review. In the current state the en-US books are kept as is to serve as a base on which to compare the content changes.

### DIFF
--- a/src/main/java/com/dreammaster/mantle/MantleBookLoader.java
+++ b/src/main/java/com/dreammaster/mantle/MantleBookLoader.java
@@ -15,14 +15,14 @@ import mantle.books.BookData;
 
 final class MantleBookLoader implements BookLoader {
 
-    static final BookDataReader BOOK_DATA_READER = new BookDataReader();
-    static final BookDataStoreProxy BOOK_DATA_STORE_PROXY = BookDataStoreProxy.getInstance();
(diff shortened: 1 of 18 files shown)


src/main/java/com/dreammaster/mantle/MantleBookLoader.java
```diff
@@ -0,0 +1,79 @@
...
+import com.google.common.base.Strings;
+
+import gregtech.GT_Mod;
+import gregtech.api.interfaces.internal.IGT_Mod;
+import mantle.books.BookData;
+
+final class MantleBookLoader implements BookLoader {
+
+    static final BookDataReader BOOK_DATA_READER = new BookDataReader();
+    static final BookDataStoreProxy BOOK_DATA_STORE_PROXY = BookDataStoreProxy.getInstance();
```
OneEyeMaker, line 19: These fields should be static, shouldn't it?

YannickMG: They are static, but they could be private.

OneEyeMaker (approved):

PR #799, opened 2024-02-08
Wooden brick form fix

chochem:
Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15495.
And brings the form in line with other gt tools with durability none of which are stackable.

Also makes the mod build again as the BS was broken.


glowredman (approved):
Making the wooden form breakable was a mistake to begin with but it is what it is.

PR #800, opened 2024-02-11
Remove lead from smeltery

chochem:
tinkers molten lead is completely useless. Being able to obtain it is just a noob trap and confuses people.

This PR removes the recipes to obtain it. Will add a WG PR as well.

closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15422


Dream-Master (approved):

Pilad:
Silver also doesn’t seem to flow into GT mechanisms. And it seems that there are several other metals with a similar bug, not just silver and lead.

Dream-Master:
not remove silver. I have plans to change alu gravel to silver to avoid this gt tiering get broken. But i had no time yet to plan a concept. Will do soon.

AnrDaemon:
Useless? What about crossbow bolts?

chochem:
useless means useless. no bolts and nothing else.

PR #808, opened 2024-02-25
Ru.lang + minor fixes in the original

kamenskyi:
Wrong string id in the brewing stand research page? and some localization mismatch - When translating, I noticed that the blood arsenal item costs the same as the blood magic item


src/main/resources/assets/dreamcraft/lang/en_US.lang
```diff
@@ -946,8 +946,8 @@ tc.research_text.DISENCHANTINGTABLE=[DC] Oh, you can remove Enchants? Nice!
...
+Minecraft.research_page.BREWINGSTAND.2=By brewing a second ingredient into a base potion in the same manner, you can create a potion with a working effect. A third ingredient may be added to make the effect more intense or last longer, or turn the effect harmful. Finally, gunpowder can be added to a potion at any stage to convert it to a splash potion, which can be thrown (or fired using a dispenser) to disperse its effect in a radius. Each brewing step takes 20 seconds. Brewing can create very useful and lifesaving items. A good number of them are useful in combat by aiding the player or weakening enemies while others can save the players life if used quickly, like Fire Resistance or Healing.
```
kuba6000, line 950: That was correct looking at the code
https://github.com/GTNewHorizons/NewHorizonsCoreMod/blob/bfb72d533633d131751781772f8fd189410845ae/src/main/java/com/dreammaster/scripts/ScriptTCCoreMod.java#L453-L454

kuba6000 (changes requested):

src/main/resources/assets/dreamcraft/lang/en_US.lang
kamenskyi: but when translated from the original it does not translate into Russian

src/main/resources/assets/dreamcraft/lang/en_US.lang
```diff
@@ -946,8 +946,8 @@ tc.research_text.DISENCHANTINGTABLE=[DC] Oh, you can remove Enchants? Nice!
...
 Minecraft.research_page.BREWINGSTAND.3=Gathering the Blaze Rods and Nether Wart necessary for brewing can prove challenging, but once they are set up, most potion ingredients are fairly plentiful and brewing will be a rewarding task.
```
kamenskyi, line 951: but why then is it standing here?

src/main/resources/assets/dreamcraft/lang/en_US.lang
kamenskyi: but why then is it standing here? Minecraft.research_page.BREWINGSTAND.3

kuba6000: > but why then is it standing here? Minecraft.research_page.BREWINGSTAND.3

I don't know, thats how it was in the old scripts.
There are curently 3 keys that the code uses:
brewingstand.research_page.BREWINGSTAND.1
brewingstand.research_page.BREWINGSTAND.2
Minecraft.research_page.BREWINGSTAND.3

We can either leave it as that or fix it in the code and all lang files.

PR #813, opened 2024-02-26
Fix battlegear hazard check

Ableytner:
Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15531.


Caedis (approved):
Looks good.
I will rework this entire method to no longer use reflection every tick

PR #815, opened 2024-02-27
Cache bg2 inventory field and add backwards compat to older BG2

Caedis:


src/main/java/com/dreammaster/modhazardousitems/HazardousItemsHandler.java
```diff
@@ -399,32 +401,28 @@ private void CheckInventoryForItems(EntityPlayer pPlayer) {
...
+                    if (bg2ExtraInvField == null) {
+                        try {
+                            bg2ExtraInvField = pPlayer.inventory.getClass().getDeclaredField("battlegear2$extraItems");
+                        } catch (NoSuchFieldException nsfe) {
+                            _mLogger.warn(
+                                    "battlegear.changed.1",
+                                    "Seems battlegear has updated/changed. Someone has to fix HazardousItems!");
+                            bg2ExtraInvField = pPlayer.inventory.getClass().getDeclaredField("extraItems");
+                        }
+                        bg2ExtraInvField.setAccessible(true);
```
OneEyeMaker, line 415: This line should be inside `try` block. And there should be null check (see line 417).

OneEyeMaker (changes requested):

src/main/java/com/dreammaster/modhazardousitems/HazardousItemsHandler.java
Caedis: It is, see line 420

Caedis: getDeclaredField can't return null, it would a throw NoSuchFieldException, which is caught

OneEyeMaker (approved):

PR #816, opened 2024-03-01
Fix moon buggy copy recipe and related fixes

chochem:
fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15611

in more detail:
- changes recipes for the copying of moon buggy/astro miner/cargo rocket schematic to be the same as the equivalent rocket tier schematic it can be exchanged with
- add programmed circuits to the recipes. otherwise they are removed as duplicates! the nbt on the datastick doesnt count. :(
- use oredict for circuits for all these recipes so that the circuits cycle in NEI instead of having 12 separate recipes for the different circuits.


HoleFish:
nbt on the data sticks should work. another issue 😱

AbdielKavash (approved):

PR #823, opened 2024-03-09
Fix broken tinkers defence recipes

chochem:
fix incorrect item names in 5 tinkers defence recipes.

fixes the last few point and thus closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15676

image
image


Ethryan (approved):
Huh, Strange naming scheme for the item ids...

chochem:
yea a bit strange for sure. I just used `/mt hand' and copied that in. :P

Dream-Master:
Use /iih         ingame is more  accurate . You get much more infos.

chochem:
its not more accurate. its the same thing on that front. and it doesnt copy.

PR #786, opened 2023-12-31
Add Apiamancer's Drainer recipe

Quantumlyy:
image
image

PR #825, opened 2024-03-17
Correct recipe mistake

Alastors:
This recipe was supposed to use tier novice cells (EV), but an oversight on my part led to it using tier adept cells (IV) which put it way out of its intended use range, this PR just changes the metadata from tier adept to tier novice


Dream-Master:
@Alastors was this changed as we moved from scripts to code?

chochem:
no, nothing like that. The recipe is from here: https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/786.

Alastors:
> @Alastors was this changed as we moved from scripts to code?

What?

Dream-Master:
was this a mistake you made as the script files converted to core mod code ?

Alastors:
> was this a mistake you made as the script files converted to core mod code ?

No this was made after that

combusterf:
Remembering a number of other recent discussions, this would effectively mean that:

- you can replicate any essentia at HV
- you can build the "trivial" set-up of the soul brazier also at HV (which is probably the most powerful warp management option short of actually removing warp)

You'll still need eldritch epiphany, and with that a good chunk of warp to unlock it. I'm skeptical about this change balance-wise, but it's not my call to make.

Dream-Master:
i freeze this for now to get more time and test stuff myself to get a better overview.

Alastors:
> i freeze this for now to get more time and test stuff myself to get a better overview.

okay, though please remember to handle this before stable, this really needs to be fixed before then if at all 😅

Alastors:
> Remembering a number of other recent discussions, this would effectively mean that:
> 
> * you can replicate any essentia at HV
> * you can build the "trivial" set-up of the soul brazier also at HV (which is probably the most powerful warp management option short of actually removing warp)
> 
> You'll still need eldritch epiphany, and with that a good chunk of warp to unlock it. I'm skeptical about this change balance-wise, but it's not my call to make.

to comment on this though, you can make the pure tear at IV which renders this basically useless when considering its speed, which basically means the bee won't ever get used.

I'm going to remind y'all, the bee is meant to be about the same tier as the vis bee, it's the same difficulty of crafting but it can't be used without this block, and because it's slow it's meant to be used like the oil bee.

Alastors:
Pinging this again.

***Reminder that this is supposed to be a more balanced patch to the hole the condenser fixed, putting it in IV just rips the entire hole open and makes this unusable***

Impos913:
From my testing it works well and can replace the condenser but it should be tiered earlier. Novice diffusion cells instead of adept cells would be proper tiering imho like HV-EV. Adept cells normally require platline, going to europa and callisto which need T3 rocket and polybenzimidazole. Currently it's gated to mid-late IV,

MeiTianyou:
My testing also shows that it works pretty well, there are some lingering issues with it interacting weirdly with the iapiary, but it's very usable. I think it should definitely be tiered before IV milestones like T3 rocket, full platline, and iv hatch.

I don't see it making soul brazier free as bad thing, given that warp is immediately a non issue when you can make pure tears (post iv hatch).  I would be relatively more concerned by it making infusion crafting more free, but we already have WA smeltery and essentia ae, so its an extremely marginal bonus on top of those. 

My suggestion is to make it tiered to EV, and (soft) put behind t2 rocket by requiring some obnoxious tungstensteel part, like two large gears or something. Additionally, I think it would be relatively more balanced if imprinting it with an essentia type was permanent, but that's a different conversation entirely.

Dream-Master (approved):

PR #828, opened 2024-03-26
let all crystal recipes require low gravity to work.

koiNoCirculation:
They're related to https://github.com/GTNewHorizons/GTplusplus/pull/861


chochem (changes requested):
I am against this. the point is to make the initial bits in space, not the entire automation.

boubou19:
nothing like this will be merged, it's just a reminder for later

Dream-Master:
yes let it open until we have a good concept which stuff have to be produced on a space base and why.

Divran:
copy pasting my comment to be sure it isn't missed

> If something is added that requires a space station then the bug that causes you to get yeeted up at mach 20 whenever you walk around on a space station needs to be fixed first. I've been told that using a copterpack is a workaround, but just because a workaround exist does not mean it isn't a bug, and we shouldn't need to carry around MV tier tech for the entire rest of the game for no reason other than to fly back to the space station every 3 seconds

koiNoCirculation:
> I am against this. the point is to make the initial bits in space, not the entire automation.

If you do not let the entire automation use low gravity, the low gravity setting is still meaningless.
You just bring a nine in one in space, craft a chip and abandon that. That's not cool.
The original aim of low gravity should be like this.
67f70c52bd2d18aba73c1c8cdbd9270f
A base connects to earth with quantum ring.
If you want making chips on earth, I suggest moving that recipe without low-gravity to UV/UHV-tier. Like what we do with cleanroom(MV-early IV).

serenibyss:
Closing this for now, will be discussed in an areas of interest post how to move forward with a low gravity rework

PR #834, opened 2024-04-12
fix mahogany recipes

chochem:
Its all done correctly by GT itself. One just has to leave it alone after that :P

fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14664


src/main/java/com/dreammaster/scripts/ScriptBiomesOPlenty.java
```diff
@@ -103,9 +103,10 @@ public void loadRecipes() {
                 getModItem(BiomesOPlenty.ID, "misc", 1, 1, missing),
                 getModItem(BiomesOPlenty.ID, "misc", 1, 1, missing),
                 getModItem(BiomesOPlenty.ID, "misc", 1, 1, missing));
-        addShapelessRecipe(
-                getModItem(BiomesOPlenty.ID, "planks", 2, 14, missing),
-                getModItem(BiomesOPlenty.ID, "logs4", 1, 3, missing));
+        /*
```
Dream-Master, line 106: if we not need this just remove it i would say.

chochem: oh yea was just testing first. forgot to remove it proper at the end. one second

Dream-Master (approved):

PR #835, opened 2024-04-13
Add missing resistor recipes

konradmoesch:
This PR fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14627 by adding the missing recipes (TC slime items + Lignite)

Marked as draft since commit 6579f42d2539cc3082ffab110fe2519ff3d1f9fe (simplify recipes by using the oredict for `slimeball`) does not work.

Is it intended that using the oredict for input items does not work? In the recipe for `iron_bars`, `stickAnyIron` (`OrePrefixes.stick, Materials.AnyIron`) is used, which is also ignored and simplified to only the default `iron rod`

Ifit is not intended to use oredict names for input items, we can revert the last commit and un-draft.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,7 +1,7 @@
 // Add your dependencies here
 
 dependencies {
-    api("com.github.GTNewHorizons:GT5-Unofficial:5.09.45.146:dev")
+    api("com.github.GTNewHorizons:GT5-Unofficial:5.09.45.148:dev")
     api("com.github.GTNewHorizons:Yamcl:0.6.0:dev")
     api("com.github.GTNewHorizons:Baubles:1.0.4:dev")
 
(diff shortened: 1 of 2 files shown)


Dream-Master:
slime is oredicted but not in gt code so thats why it not work

Dream-Master (approved):

konradmoesch:
Thanks for merging!

Is it intended that oredicted but non-gt items do not work in the recipes? I feel that it would be nice if all items, also non-gt, could be used in a unified way.

chochem:
you can absolutely use oredict for crafting recipes. But GT_OreDictUnificator.get does not do anything like that.

The script files in this mod should have many examples for you to look at for crafting recipes with oredicts.

PR #836, opened 2024-04-19
Unstable ingot fixes

chochem:
fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14075
actually fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/5416

Fixes several aspects of unstable ingot crafting:
- couldnt craft the mobius ingot at all. one would get the regular unstable ingot instead
- could craft the regular unstable ingot without activating the sigil
- the shape of the recipe was wrong

Does NOT fix: stack crafting bypasses the explosion mechanic. (idea: maybe we can overwrite the maximum stacksize and set it to 1?)

image
activated sigil: correctly uses durability and correctly gets the 10s cooldown to explosion (if crafted one at a time)

image
pseudo-inversion sigil: does not use durability and makes the mobius ingot.

### DIFF
--- a/src/main/java/com/dreammaster/recipes/CustomItem.java
+++ b/src/main/java/com/dreammaster/recipes/CustomItem.java
@@ -56,6 +56,8 @@ public interface CustomItem {
         private NBTTagCompound nbt;
         private boolean exact = false;
 
+        private boolean noValues = false;
+
         public NBTItem() {}
 
         public NBTItem(ItemStack stack) {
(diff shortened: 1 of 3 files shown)


Ethryan:
I think changing the stacksize for the exploding ones is probably the easiest way of fixing the issue.

Ethryan:
Ah, I forgot about that.

glowredman:
> I am not sure how to do that though (in particular since the mod is ARR) :P

Get the instance of the Item (e.g. with `GameRegistry.findItem()` or some random GT method) and call `setMaxStackSize()` on it.

chochem:
to quote myself from discord: it looks like this wouldnt work anyway. if you 'shift' craft even a single one, it doesnt explode.
This probably needs a fix whereever we did the shiftcraft optimizations.

I am happy with this PR as is for now.

Ethryan (approved):

PR #844, opened 2024-05-08
Fix planet glowstone recipe loop.

AbdielKavash:
It is currently possible to craft planet-specific glowstone from GalaxySpace using 1 glowstone dust and 1 [planet] stone dust to make 2 [planet] glowstone dusts. One [planet] glowstone dust can also be centrifuged into 2 glowstone dusts and a chance-based output of [planet] stone dusts. This allows for infinite creation of glowstone and planet stone.

This PR changes the [planet] glowstone dust recipes to take 2 glowstone dusts and 2 [planet] stone dusts to craft 1 [planet] glowstone dust. This means that centrifuging the [planet] glowstone dust returns all of the glowstone, and has a chance to return the [planet] stone dust used. It is no longer resource positive.

Thanks to **@\_pix3lated\_** on discord for reporting this.


chochem (approved):
makes sense

PR #840, opened 2024-04-27
Project Red Pipe rework

Dream-Master:
I rework Project Red pipes so they all doable in LV 

```
+ 1 iron rod instead of 2 for Item Transport Pipe when handcrafted
+ 2 iron rod instead of 4 for Item Transport Pipe when assembling machine crafting (and double throughput/s)
+ New more efficient assembler tiers for Item Transport pipe using steel, aluminium
+ Removed infused silicon entirely in Junction Pipe recipe
+ Metal part of 4x Junction Pipe recipe is 1 steel gear + 2 steel rods instead of 1 aluminium rod
+ Junction Pipe recipe gets cheaper in assembler (less steel)
+ New aluminium, stainless, and titanium tiers for Junction Pipe assembler recipe
+ Interface Pipe gets a new handcraft recipe (4 iron screw + 2 gold plate + 1 Junction Pipe + 1 circuit board + 1 redstone plate)
+ base Interface Pipe in assembler recipe costs an additional 1 circuit board (but has 2x throughput)
+ Interface Pipe now has circuit tiering, allowing you to do up to 64 Junction Pipe -> 64 Interface Pipe (if you spend wetwares)
+ Routed Request Pipe requires a circuit now, but diamond plate and redstone cost is reduced
+ New efficient circuit tiers for chips (Null Logic chips use ULV circuits chips and chipkards Lv circuits)
```

Screenshot 2024-04-26 12 11 38
Screenshot 2024-04-26 12 12 03
Screenshot 2024-04-26 12 12 13
Screenshot 2024-04-26 12 12 38
Screenshot 2024-04-26 12 12 52
Screenshot 2024-04-26 12 13 01
image
image

PR #845, opened 2024-05-09
Restore (accidentally?) removed ProjectRed recipes.

AbdielKavash:
In https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/840, Project Red Transportation has been downtiered and many recipes changed. However, in this PR, recipes for several PR parts have been completely removed: Silicon, Red Silicon Compound, Infused Silicon, and Silicon Chip. These recipes make up a crafting chain, and the Silicon Chip is used to craft several PR logic gates, which are unobtainable otherwise. This PR restores the four removed recipes.

We talked about this with @chochem on discord, and I believe this removal was accidental, as I see no reason for it. I suspect @Dream-Master simply did not notice that the chips have other uses than logistic pipes and thought them redundant. If there is any legitimate reason to remove these recipes, then an alternate way of crafting the affected gates should be added instead.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,12 +1,12 @@
 // Add your dependencies here
 
 dependencies {
-    api("com.github.GTNewHorizons:GT5-Unofficial:5.09.46.05:dev")
+    api("com.github.GTNewHorizons:GT5-Unofficial:5.09.46.09:dev")
     api("com.github.GTNewHorizons:Yamcl:0.6.0:dev")
     api("com.github.GTNewHorizons:Baubles:1.0.4:dev")
 
     compileOnly("com.github.GTNewHorizons:Avaritia:1.49:dev") { transitive = false }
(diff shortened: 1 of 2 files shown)


chochem (dismissed):
looks good.

One use was removed (on purpose). So yea, dream probably didnt realize they have other uses.

Dream-Master:
I will check tomorrow . Please not merge yet. I was thinking to use gt silicon boule / silicon chips from gt instead when i change recipes but bouls are late lv. And maybe a engraver to get different colors . (Yes I not realized they used elsewhere )

AbdielKavash:
Sounds good, we already suggested some alternate recipes. I don't mind, as long as the gates are craftable again <3

Dream-Master:
image
image
image
image
image
image
image
image
image

OrderedSet86 (approved):
Looks good to me

PR #860, opened 2024-06-03
fix GT++ drops not being added to deepdark

boubou19:


Alastors (approved):
I don’t personally see anything wrong here

PR #867, opened 2024-06-09
Downtier HP steam turbine to EV

HoleFish:
TungstenSteel -> titanium, LuV circuits -> IV. 
IV machine hull stays the same. So the tungstensteel requirement is reduced from 36 to 8.
See also https://github.com/GTNewHorizons/GT5-Unofficial/pull/2647


Dream-Master (changes requested):
No why ?

HoleFish:
See https://discord.com/channels/181078474394566657/401118216228831252/1249341830563106908

Dream-Master (approved):

PR #875, opened 2024-06-20
Allow converting circuits into the Any circuit

S4mpsa:
We can already do this with the circuit wraps, why not with normal circuits?

This PR adds recipes in the Forming Press to turn any tiered circuit into the corresponding Any circuit of the tier. Such change would simplify changing the circuit type used in AE2 patterns significantly.

image

https://github.com/GTNewHorizons/GT5-Unofficial/pull/2682 fixes LV and HV circuits not working properly.

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/FormingPressRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/FormingPressRecipes.java
@@ -334,31 +334,31 @@ public class FormingPressRecipes implements Runnable {
                 .itemOutputs(new ItemStack(Blocks.glass, 1, 0)).duration(5 * SECONDS).eut(TierEU.RECIPE_LV)
                 .addTo(formingPressRecipes);
 
-        // Recipes for Any circuits
+        // Recipes for Any circuits (LV and HV are broken due to IC2 interactions)
         GT_Values.RA.stdBuilder()
                 .itemInputs(
                         GT_OreDictUnificator.get(OrePrefixes.circuit, Materials.Primitive, 1),
                         GT_Utility.getIntegratedCircuit(24))
                 .itemOutputs(com.dreammaster.item.ItemList.CircuitULV.getIS(1)).duration(1 * SECONDS)
                 .eut(TierEU.RECIPE_LV).addTo(formingPressRecipes);
-        GT_Values.RA.stdBuilder()
-                .itemInputs(
-                        GT_OreDictUnificator.get(OrePrefixes.circuit, Materials.Basic, 1),
-                        GT_Utility.getIntegratedCircuit(24))
-                .itemOutputs(com.dreammaster.item.ItemList.CircuitLV.getIS(1)).duration(1 * SECONDS)
-                .eut(TierEU.RECIPE_LV).addTo(formingPressRecipes);
+        //GT_Values.RA.stdBuilder()
+        //        .itemInputs(
+        //                GT_OreDictUnificator.get(OrePrefixes.circuit, Materials.Basic, 1),
+        //                GT_Utility.getIntegratedCircuit(24))
+        //        .itemOutputs(com.dreammaster.item.ItemList.CircuitLV.getIS(1)).duration(1 * SECONDS)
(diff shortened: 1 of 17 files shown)


src/main/java/com/dreammaster/gthandler/recipes/FormingPressRecipes.java
```diff
@@ -334,6 +334,105 @@ public void run() {
                 .itemOutputs(new ItemStack(Blocks.glass, 1, 0)).duration(5 * SECONDS).eut(TierEU.RECIPE_LV)
                 .addTo(formingPressRecipes);
 
+        // Recipes for Any circuits (LV and HV are broken due to IC2 interactions)
+        GT_Values.RA.stdBuilder()
+                .itemInputs(
+                        GT_OreDictUnificator.get(OrePrefixes.circuit, Materials.Primitive, 1),
+                        GT_Utility.getIntegratedCircuit(24))
+                .itemOutputs(com.dreammaster.item.ItemList.CircuitULV.getIS(1)).duration(1 * SECONDS)
```
HoleFish, line 342: `.itemOutputs(GT_OreDictUnificator.get(OrePrefixes.circuit, Materials.something, 1))` works fine and is compatible even for LV and HV circuits (return IC2 circuits). I suggest you use it for god's sake

chochem: you want a specific output, so why would you use unificator? Just put in the output you want.

HoleFish: You can use unificator to get a specific itemstack, no contradiction. And it's compatible even if we change circuit oredict (like in https://github.com/GTNewHorizons/GT5-Unofficial/pull/2675) so it will be safe

chochem: This is not about contradictions.
For inputs unificator is fine or even positively good because they are unified, so it makes the code easier to read. But outputs are not unified obviously but a specific item. So you dont want to be vague about what item it is.

HoleFish: ok then. I change my suggestion: if you failed to add oredict for IC2 circuits, add IC2 circuits in output directly because they are the "specific items for unification"

S4mpsa: Changing the output to IC2 circuits is thematically weird but mechanically the best option. I'll do that so we have full range of circuits, even if LV and HV are named differently.

S4mpsa: > `.itemOutputs(GT_OreDictUnificator.get(OrePrefixes.circuit, Materials.something, 1))` works fine and is compatible even for LV and HV circuits (return IC2 circuits). I suggest you use it for god's sake

I indeed had it like this initially, and it worked fine, but changed them to use specific items to attempt to fix the LV/HV circuit issue. I could change them back, but it's not immediately obvious that we want to convert all circuit variants into a single variant from that code - after all, it just returns *a circuit* of the appropriate tier. That circuit just happens to be the Any circuit by pure chance.

HoleFish: That's not "happens to be". That's how recipe matching works

S4mpsa: Why does it match to the Any circuit in this case?
Does the `ModSimpleBaseItem` make any items made with it to be the default when oreDict is used to unify? It's not immediately clear to me why the Any circuit is chosen by the unificator from the 6+ possibilities. I genuinely don't know how it's chosen.

chochem:
with https://github.com/GTNewHorizons/GT5-Unofficial/pull/2682 you can use the anycircuits for LV and HV

chochem (approved):
looks good. I am sure many people will appreciate it.

But don't merge without merging the mentioned GT PR.

PR #877, opened 2024-06-24
Fix many circuit oredicts in nhcoremod

chochem:
this is a good code cleanup change on its own.
But additionally, it is needed for https://github.com/GTNewHorizons/GT5-Unofficial/pull/2682.


S4mpsa (approved):
Some recipes are still a bit weird, having only ore input cleaned up. Still a good cleanup.

PR #885, opened 2024-07-03
Fix like this since, enums.Mods don't contain the value for akashic tome yet.

Ethryan:
Now it works


Dream-Master (approved):

chochem:
to fast dream, they just added the enum

PR #897, opened 2024-07-18
Wireless data recipes

NotAPenguin0:
All recipes tiered at UMV.

image
image

image
image


EnderProyects:
Just saying but for parity shouldnt all crafts use the internet card?

NotAPenguin0:
Yeah this makes sense honestly, I don't really know why I didn't make the data stick ones use it

EnderProyects:
I mean in the description all are internet conected xd

S4mpsa (approved):
Tested in full pack and everything looks as expected.

One problem for the future might be the explicit usage of the Quantum circuit, if other UXV circuits are ever added. But this should be fixed by adding oreDict to the Quantum circuit at that point.

PR #900, opened 2024-07-19
Fix UMV and UXV circuit oreDicts

S4mpsa:
The circuits have been oreDicted to the wrong Material enum since forever. With some new UMV and UXV circuits (namely the Any circuits), this actually starts to matter.

Also changes a few recipes which were using non-oreDicted circuits in their recipes.

Dependancy of:
https://github.com/GTNewHorizons/GTNH-Intergalactic/pull/64
https://github.com/GTNewHorizons/GT5-Unofficial/pull/2739

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,7 +1,7 @@
 // Add your dependencies here
 
 dependencies {
-    api("com.github.GTNewHorizons:GT5-Unofficial:5.09.48.93:dev")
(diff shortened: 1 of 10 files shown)


S4mpsa:
The build fails because of a badly-implemented oreDict ItemStack getter in gt5u which causes a NPE error when the oreDict doesn't have anything in it, which happens to be the case after this PR as it moves the circuit oreDicts. Fixed by https://github.com/GTNewHorizons/GT5-Unofficial/pull/2739

boubou19 (approved):

PR #902, opened 2024-07-20
Add new recipes to star fuel and dark matter

GDCloudstrike:
This PR adds a new recipes to the avaritia star fuel item and amun ra dark matter, both of which are planned to be used in the upcoming godforge multiblock. Dark matter can be found naturally generating on the mehen belt, these recipes are optional ones for those who dont want to go search for it in world.
image
image
image

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/DTPFRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/DTPFRecipes.java
@@ -16,6 +16,7 @@ import com.dreammaster.gthandler.CustomItemList;
 import com.dreammaster.gthandler.DTPFCalculator;
 import com.dreammaster.gthandler.GT_CoreModSupport;
 
+import goodgenerator.items.MyMaterial;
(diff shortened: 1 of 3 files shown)


NotAPenguin0 (approved):
Using mk5 naqfuel in the recipe is evil, I like it.

Reflex18:
> Using mk5 naqfuel in the recipe is evil, I like it.

Doesnt that basically enforce one power type over the other?

NotAPenguin0:
> Doesnt that basically enforce one power type over the other?

It's only used for crafting, so not really.

Persus1:
I dont know how much Dark Matter you will need, but i do remember that you could use Fortune on the Dark Matter blocks. With this you could get over a stack of Dark Matter with the World Breaker(or more if you create even higher Fortune Levels). So if you dont search on Mehen Belt for the really rare meteor you probably just need to do this recipe one time.

Just thought about this while seeing the Extractor recipe for just 8 pieces.

PR #907, opened 2024-07-24
Add universe fluid cell recipe

EnderProyects:
Added a GTNH recipe for the Fluid extreme universe cell

related
https://github.com/GTNewHorizons/AE2FluidCraft-Rework/pull/218

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/SpaceAssemblerRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/SpaceAssemblerRecipes.java
@@ -6,28 +6,30 @@ import static gregtech.api.enums.Mods.BartWorks;
 import static gregtech.api.enums.Mods.GTNHIntergalactic;
 import static gregtech.api.enums.Mods.GTPlusPlus;
 import static gregtech.api.enums.Mods.GoodGenerator;
+import static gregtech.api.enums.Mods.KekzTech;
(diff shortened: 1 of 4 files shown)


src/main/java/com/dreammaster/gthandler/recipes/SpaceAssemblerRecipes.java
```diff
@@ -12,23 +12,22 @@
...
-import net.minecraft.nbt.NBTTagCompound;
-import net.minecraftforge.fluids.Fluid;
-import net.minecraftforge.fluids.FluidRegistry;
-import net.minecraftforge.fluids.FluidStack;
-
+import com.dreammaster.item.ItemList;
 import com.github.technus.tectech.thing.CustomItemList;
 import com.gtnewhorizons.gtnhintergalactic.recipe.IG_RecipeAdder;
 
-import gregtech.api.enums.ItemList;
```
S4mpsa, line 24: You removed the ItemList from GT5u and included the ItemList from CoreMod - yes these should have separate names but this means you cannot access the item enums from GT5u.

You can circumvent this by using the full name of either list (I suggest CoreMod, as the recipe code already does this, which makes me think your IDE pulled a prank on you.)

Example:
`com.dreammaster.item.ItemList.CircuitUXV.getIS(4)`

Dream-Master (approved):

PR #913, opened 2024-07-31
Fix for Issue #16879

AlexanderChenhall:
Added check for if the game is currently in Peaceful, and if so, to cancel the damage event pre-emptively.
For reason, see [Issue #16879](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/16879)

### DIFF
--- a/src/main/java/com/dreammaster/baubles/OvenGlove.java
+++ b/src/main/java/com/dreammaster/baubles/OvenGlove.java
@@ -12,6 +12,7 @@ import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.util.IIcon;
+import net.minecraft.world.EnumDifficulty;
 import net.minecraft.world.World;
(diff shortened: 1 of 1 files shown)


src/main/java/com/dreammaster/baubles/OvenGlove.java
```diff
@@ -210,6 +211,12 @@ public void onDamage(LivingAttackEvent event) {
             }
             EntityPlayer player = (EntityPlayer) entity;
 
+            // If the game is in peaceful, exit early. Why do this? See issue #16879
```
Caedis, line 214: Move to after the `DamageSourceHotItem` check

Caedis (dismissed):

src/main/java/com/dreammaster/baubles/OvenGlove.java
AlexanderChenhall: Damn, done. Not sure how i missed that my bad.

boubou19 (approved):

PR #918, opened 2024-08-04
Do not rely on class load order in BioItem loader

eigenraven:
Split BioItemLoader item registration to run earlier, and deterministically via a method instead of at class load time.


Dream-Master (approved):
fixed my issue

PR #919, opened 2024-08-04
Battery tier fixes

chochem:
Follow up to https://github.com/GTNewHorizons/GT5-Unofficial/pull/2825.
Fixes battery tiers in various recipes to be in line with new tier materials.

For some cursed reasons the EV lapo battery was tagged 'master' in the past (which would mean LuV) and the energy crystal 'elite' (which would mean IV) despite being HV.


Pxx500 (approved):
yeah the naming's weird

PR #928, opened 2024-08-10
Added wormhole generator recipe

RecursivePineapple:


src/main/java/com/dreammaster/gthandler/recipes/AssemblingLineRecipes.java
```diff
@@ -1125,5 +1126,24 @@ public void run() {
                 20000,
                 (int) TierEU.RECIPE_UMV);
```
S4mpsa, line 1128: Add a comment detailing what this recipe is adding, as is done with all other recipes in this file

S4mpsa (approved):

PR #930, opened 2024-08-12
Don't classload GC when it's not present

ah-OOG-ah:
Moves some of the Galacticraft script into it's own file, because it directly references Galacticraft and as such can't be classloaded if Galacticraft isn't present


kuba6000 (changes requested):
You can just put `@Optional.Method(modid = "GalacticraftCore")` on `astroMinerRecipes` ([like that](https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/934/commits/f5fd079e58d52f8d5c21a1ad4f5bae3098ba2bbd)), there is no need to move everything to another class

kuba6000 (approved):

PR #936, opened 2024-08-16
fix catwalks recycling recipes

chochem:
fixes infinite steel exploit, fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/17021, and turns the recipes into proper recycling recipes.


boubou19:
mhm, maybe we should deprecate this in favor of RA2, but anyways, LGTM

boubou19 (approved):

chochem:
there is no ra2 for itemdata or anything like that.

boubou19:
yet... :D

PR #941, opened 2024-08-21
Remove one thousand (!) ztones collisions

FourIsTheNumber:
We were previously trying to generate recipes for every ZTones block that uses dye with every oredicted dye variant. This did not actually work, and caused broken recipes that looked like this (and about a thousand recipe collisions):
image

I have made it not try to use oredicted dye and just use the GT dye instead, since actually trying to fix the oredicting would just be adding thousands of bloated recipes to assembler.

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
@@ -5924,13 +5924,13 @@ public class AssemblerRecipes implements Runnable {
 
         final int ztoneVariants = 16;
 
+        final ItemStack ztoneStone = GT_ModHandler.getModItem(ZTones.ID, "stoneTile", 4L, 0);
+        final ItemStack ztoneAurora = GT_ModHandler.getModItem(ZTones.ID, "auroraBlock", 4L, 0);
+
         for (String name : blockName) {
             for (int i = 0; i < ztoneVariants; i++) {
                 GT_Values.RA.stdBuilder()
-                        .itemInputs(
-                                GT_ModHandler.getModItem(ZTones.ID, "stoneTile", 4L, 0),
-                                itemMap.get(name),
-                                GT_Utility.getIntegratedCircuit(i == 0 ? 24 : i))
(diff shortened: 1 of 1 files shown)


src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
```diff
@@ -5888,68 +5886,64 @@ private void makeZTonesRecipes() {
...
-                            .itemInputs(
-                                    GT_ModHandler.getModItem(ZTones.ID, "stoneTile", 4L, 0),
-                                    ingredient,
-                                    GT_Utility.getIntegratedCircuit(i == 0 ? 24 : i))
-                            .itemOutputs(GT_ModHandler.getModItem(ZTones.ID, "tile." + name + "Block", 8L, i))
-                            .duration(10 * SECONDS).eut(16).addTo(assemblerRecipes);
-                }
+                GT_Values.RA.stdBuilder()
+                        .itemInputs(
+                                GT_ModHandler.getModItem(ZTones.ID, "stoneTile", 4L, 0),
```
boubou19, line 5931: this could be cached outside the double for loop and passed in the item inputs

src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
```diff
@@ -5888,68 +5886,64 @@ private void makeZTonesRecipes() {
...
-                            .itemInputs(
-                                    GT_ModHandler.getModItem(ZTones.ID, "auroraBlock", 4L, 0),
-                                    ingredient,
-                                    GT_Utility.getIntegratedCircuit(i == 0 ? 24 : i))
-                            .itemOutputs(GT_ModHandler.getModItem(ZTones.ID, "tile." + name + "Block", 8L, i))
-                            .duration(10 * SECONDS).eut(16).addTo(assemblerRecipes);
-                }
+                GT_Values.RA.stdBuilder()
+                        .itemInputs(
+                                GT_ModHandler.getModItem(ZTones.ID, "auroraBlock", 4L, 0),
```
boubou19, line 5942: this could be cached outside the double for loop and passed in the item inputs

boubou19 (changes requested):

boubou19 (approved):

PR #944, opened 2024-08-25
Move galactic greg space ore registration to GT5U

boubou19:
Now the mergening happened, there is no reason to register space ores through NHCore, especially since it's doing reflection to do this.

Depends on https://github.com/GTNewHorizons/GT5-Unofficial/pull/2939


Caedis (approved):
:shipit:

PR #956, opened 2024-08-30
Fix quicksilver singularity

FourIsTheNumber:
Quicksilver singularity was broken because there is no properly gt-oredicted quicksilver block. Just made it cost 9x the amount in cinnabar dust instead.


TheEpicGamer274 (approved):
Looks fine to me

PR #962, opened 2024-08-31
Reintroduce titanium loop loss

Pxx500:
Reintroduces titanium loop loss which was deleted as a result of the chemical rebalance™️

Instead of the old loss (50% of chlorine lost per ingot), this loss will be 12,5% loss of chlorine and magnesium per ingot.

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/BlastFurnaceRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/BlastFurnaceRecipes.java
@@ -1029,7 +1029,7 @@ public class BlastFurnaceRecipes implements Runnable {
                         GT_Utility.getIntegratedCircuit(12))
                 .itemOutputs(
                         GT_OreDictUnificator.get(OrePrefixes.ingotHot, Materials.Titanium, 1L),
-                        GT_OreDictUnificator.get(OrePrefixes.dust, Materials.Magnesiumchloride, 6L))
+                        GT_OreDictUnificator.get(OrePrefixes.dust, Materials.Magnesiumchloride, 5L))
                 .fluidInputs(Materials.Titaniumtetrachloride.getFluid(1000L)).duration(40 * SECONDS)
                 .eut(TierEU.RECIPE_HV).metadata(COIL_HEAT, 2140).addTo(blastFurnaceRecipes);
 


querns:
Was the magnesium ever actually lost before? I don't think that it was; just the chlorine.

Pxx500:
> Was the magnesium ever actually lost before? I don't think that it was; just the chlorine.

Nope, only chlorine, but adding the loss in the EBF makes most sense, also magnesium is a barely used material so loss on it won't impact anything in a meaningful way.

FourIsTheNumber:
I think one of the first major passive chemical lines being loopable is a positive for teaching players. Just seems like a pointlessly irritating change since chlorine is already spam consumed in every line in the game

chochem:
> I think one of the first major passive chemical lines being loopable is a positive for teaching players.

the first chemline is silicon which is indeed lossless. this is more than a tier later.

a quarter of the old loss in chlorine in return for some slight magnesium loss sounds good to me.

Ruling-0:
I really don't get the balance or design reasoning that necessitates this change. Setting up my first actual processing setup (LV silicon loop _barely_ counts as a line) was fun, and even encouraged me to setup bauxite proc. Setting things up to reroute looped materials was pretty interesting.

The loss in something like platline has design merit because that leads the player to setting up sources of all the basic chemicals and gasses, _when they have the tools to do so_ and will require them later. I don't see how this is more fun or more interesting design, since the solution later is trivial, if you even bother using rutile instead of just burning the directly-acquirable titanium dust, and early on I'd just slap a drawer of magnesium and salt onto the setup.

I'm actually not sure how much magnesium I had in EV, given I only have <100k in my LuV base. Chlorine has a lot of passive sources since it's consumed in large amounts everywhere. On looking, a good source is actually ilmenite slag, which meets the new 1 magnesium per 9 rutile, so I guess we just make Ilmenite even more attractive than it already was?

boubou19:
why the loss is required again? Because it seems to be a nerf for the sake of it. I get that the loss was gone, but it was a good thing, too much pressure on the chlorine.

Dream-Master:
i vote for no. Why ? is there any actual reason to change the recipe ?

Pxx500:
I've made the PR mostly because the deletion of the loss was a byproduct rather then the goal of chemical rebalance, so I thought that it should be brought back.

But since it's considered bad I'll just close the PR

Dream-Master:
i dont see a real reason. If there is some rework which make the process better/more logic i am not against it but only change this value make no sense for me.

PR #967, opened 2024-09-04
Rework indium comb processing

StaffiX:
Indium bee is possibly the strongest bee by far, as its comb solves an otherwise difficult to get material until much later, by being able to directly process its comb to indium once the player gets to LuV.
This and the related pr (https://github.com/GTNewHorizons/GT5-Unofficial/pull/3051) tries to balance it by removing the direct processing recipe, and instead using the comb to boost existing indium sources.

This pr affects:
Indium Concentrate
Milling Processes
Naquadria line
GoodGen Thorium Fuel processing
Crushed indium to purified indium

Indium comb boosts those output by 3x, and each additional indium uses 2 combs.

The related pr breaks indium phosphate processing by giving it a circuit and making LCR only.

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/ChemicalReactorRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/ChemicalReactorRecipes.java
@@ -597,6 +597,7 @@ public class ChemicalReactorRecipes implements Runnable {
                         GTOreDictUnificator.get(OrePrefixes.dust, Materials.Indium, 6),
                         GTOreDictUnificator.get(OrePrefixes.dust, Materials.TricalciumPhosphate, 5))
                 .duration(1 * SECONDS).eut(TierEU.RECIPE_LV).addTo(multiblockChemicalReactorRecipes);
+
         GTValues.RA.stdBuilder()
                 .itemInputs(
                         GTOreDictUnificator.get(OrePrefixes.crushed, Materials.Indium, 4),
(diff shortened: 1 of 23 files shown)


src/main/java/com/dreammaster/gthandler/recipes/ChemicalReactorRecipes.java
```diff
@@ -572,6 +576,34 @@ public static void multiBlockOnly() {
...
+                .itemInputs(
+                        indiumPhosphate.get(OrePrefixes.dust, 12),
+                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Calcium, 3),
+                        GTBees.combs.getStackForType(CombType.INDIUM, 8),
+                        GTUtility.getIntegratedCircuit(2))
+                .itemOutputs(
+                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Indium, 6),
+                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.TricalciumPhosphate, 5))
+                .duration(1 * SECONDS).eut(TierEU.RECIPE_LV).addTo(multiblockChemicalReactorRecipes);
+        GTValues.RA.stdBuilder()
```
kstvr32, line 600: nit: Missing space between recipes

kstvr32:
Seems reasonable to me. The Indium bee is way too strong and trivializes a previously hard to get material. 

This should incentive players to invest in the milling/frothing line until they can get Space Elevator.

boubou19 (approved):

dimazha:
Looks interesting, finally Vacuum Furnace will revive. But I've seen recipes of Pine oil – there are two: from normal steam and from SH steam
image
Maybe we need an option to produce SH steam not only with hot liquids? Maybe add a recipe of SH steam for multiblock Fluid Heater? Now it is used only for wetware and bio fluids, there will be a new option for this multi.

PR #968, opened 2024-09-05
Added large fluid extractor & solenoid recipes

RecursivePineapple:
Large Fluid Extractor (EV tiered):

Each solenoid tier made in an AL needs the previous tier as its research item.

MV Solenoid:

HV Solenoid:

EV Solenoid:

IV Solenoid:

LuV Solenoid:

ZPM Solenoid:

UV Solenoid:

UHV Solenoid:

UEV Solenoid:

UIV Solenoid:

UMV Solenoid:


RecursivePineapple:
I just realized I forgot to made sure there aren't any conflicts :pensib:
1 sec

RecursivePineapple:
Yep, the assembling machine recipes are fine. AFAIK you can't get conflicts in the AL without using the exact same items in a recipe, but let me know if I should check those too.

boubou19:
if an AL has a recipe with inputs that is a subset of another recipe input in the same order from the slice 1, the recipe that is the subset of the other can happen, depending on the order of datastick checks.

RecursivePineapple:
It looks like the AL recipes should be good. There aren't any recipes that use 2x superconductors as their first ingredient.

Dream-Master (approved):

PR #1004, opened 2024-09-18
Meteor miner recipe

zTotto:
I added the recipes for the meteor miner schematics as asked by DreamMaster:
T1 Schematic -> T6 Rocket Schematic Chip + ZPM circuit + Indalloy
T2 schematic -> T7 Rocket Schematic Chip + UV circuit + Indalloy
Both are Engraving recipes

For rocketless recipes:
T1 Schematic -> 32 Wetware Circuit Board + Quantum Anomaly + 4 Laser Emitter + Data Orb
T2 Schematic -> 32 Bio Circuit Board + Quantum Anomaly + T1 Schematic + Data Orb
Both are UV recipes in the circuit assembler

The build will fail since there's no schematic in current GT5 stable version
* https://github.com/GTNewHorizons/GT5-Unofficial/pull/3116

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/GT_CraftingRecipeLoader.java
+++ b/src/main/java/com/dreammaster/gthandler/GT_CraftingRecipeLoader.java
@@ -1,5 +1,6 @@
 package com.dreammaster.gthandler;
 
+import static bartworks.common.loaders.ItemRegistry.bw_realglas;
 import static gregtech.api.enums.Mods.AE2FluidCraft;
 import static gregtech.api.enums.Mods.AdventureBackpack;
 import static gregtech.api.enums.Mods.AppliedEnergistics2;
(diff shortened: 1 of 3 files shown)


FourIsTheNumber:
I do not think this should have a separate set of recipes for no rocket - that is extremely unusual for a multi to have an entire second recipe set. This multiblock already obsoletes a bunch of more complicated setups (including one of the remaining cool pieces of magic progression) with a simple standardized gt multi, at least let these setups thrive in niche environments like challenge runs if we're going to kill them from the standard pack.

This multiblock is also a massive buff to meteors (which are already very strong and crucial to no rocket) so it comes off as an immense targeted buff towards a challenge run.

Dream-Master:
I ask for two recipes a uv one and a T6/T7 one if you  use rockets . @FourIsTheNumber if you have a better idea for tiering just wrote it here .

EnderProyects:
> I do not think this should have a separate set of recipes for no rocket - that is extremely unusual for a multi to have an entire second recipe set. This multiblock already obsoletes a bunch of more complicated setups (including one of the remaining cool pieces of magic progression) with a simple standardized gt multi, at least let these setups thrive in niche environments like challenge runs if we're going to kill them from the standard pack.
> 
> This multiblock is also a massive buff to meteors (which are already very strong and crucial to no rocket) so it comes off as an immense targeted buff towards a challenge run.

You just obsoltete arcane bores, ( that could be used in the recipe of the controller) but is a multi for minning meteors, one thing only avalible doing blood magic, and you need thaumcraft to do blood magic, so basically you dont apart magic at one side

FourIsTheNumber:
> You just obsoltete arcane bores, ( that could be used in the recipe of the controller) but is a multi for minning meteors, one thing only avalible doing blood magic, and you need thaumcraft to do blood magic, so basically you dont apart magic at one side

It's not simply about not using the arcane bore item, its about the fact that the bore setup is much more complicated and interesting and this replaces it with a (even stronger) simple input-output gt multi for no effort.

arabadullah:
> If you like and love lag you could stay using arcane bores

So true king this is why every mechanic should be replaced with a tps efficient magic box that Does It For You™️

FourIsTheNumber:
Horrible excuse for removing complexity in favor of coded solutions

Caedis:
The bare minimum is fortune being completely removed.

EnderProyects:
I love thaumcraft but I dont like lag, is not that complicated to understand, also the "complicated setup" of arcane bores you say is just placing a grid of pipes, chests, arcane bores and vis relays, is not that complicated

Alastors:
We could make a fancy setup that isn't laggy instead of this sort of abomination, it's a bit boxy in execution, no?

EnderProyects:
@FourIsTheNumber since you are the multiblock dev, you could come up with a mechanic that is integrated in this multi so it needs for exaple magic blood blood per minning operation to work, that also consumes vis, or that you need to insert somehow essentia into it and balance determinated aspects using some redstone logic and filtered pipes ( like balancing the primordial aspects to they are always the same amount or similar)

zTotto:
I honestly found bores quite boring and annoying, there's nothing interesting in setting them up to mine meteors, you just space them evenly and use one of the 100 methods to move items. I made this multi only to have a more compact and less annoying method to mine meteors.
The fortune stuff can be removed or reworked, i have no problems in that regard (more coding practice ahah),

PR #1010, opened 2024-09-23
Buff compressor casings

FourIsTheNumber:
Related to https://github.com/GTNewHorizons/GT5-Unofficial/pull/3261

Again, hoping I can push this through since it's pretty egregious and related to beta feedback 🙏 

Reduces the incoloy-903 cost of the compressor casings by a lot, replacing some with titanium. Frankly, this multi is still very expensive at its tier. It was way way too much before and that's my fault for not really thinking about the EV cost when I was making these recipes.

Also made the stabilization casings for neutronium compressors give 4 (from 2) per recipe, since their cost is very high in field generators and the multi is already much more expensive than its original singleblock.

image
image
image

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/GT_CraftingRecipeLoader.java
+++ b/src/main/java/com/dreammaster/gthandler/GT_CraftingRecipeLoader.java
@@ -733,22 +733,22 @@ public class GT_CraftingRecipeLoader extends gregtech.loaders.postload.CraftingR
 
         // Electric Compressor Casing
         GTModHandler.addCraftingRecipe(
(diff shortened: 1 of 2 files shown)


Ethryan:
Giving 2 Compressor Casings from covering 1 Frame with 6 plates feels so wrong in my opinion.

FourIsTheNumber:
> Giving 2 Compressor Casings from covering 1 Frame with 6 plates feels so wrong in my opinion.

That's fair tbh. Maybe I can just change out some of the plates, the multi is unfortunately quite large lol. Give me a minute

serenibyss (approved):

PR #1023, opened 2024-10-08
Add ichorium block compressor recipe

StaffiX:
Adds an UV compressor recipe, using UU Matter to make it more expensive. This recipe also uses HIP make it more gated.
image


EnderProyects:
If you are going for this, make that at least it needs one of the two new compresor mechanics, the heat one or black hole

StaffiX:
Downtiered it, changed it to take UU Matter and made it use HIP.

Dream-Master:
didnt see it use HIP Unit

Dream-Master (approved):

PR #1024, opened 2024-10-09
Added `/wiki` chat command.

AbdielKavash:
The `/wiki` chat command allows the player to instantly access the GTNH wiki from the game, and search for specific topics (for example, information about blocks or game features).

Planned options:

- [x] `/wiki` - Opens the wiki main page.
- [x] `/wiki <topic>` - Searches the wiki for the specified topic. For example, `/wiki EBF` takes you to https://wiki.gtnewhorizons.com/wiki/Electric_Blast_Furnace.
- [ ] `/wiki hand` - Searches the wiki for the item currently held in the player's hand.
- [ ] `/wiki block` - Searches the wiki for the block the player is currently looking at.

First two variants are already implemented; the other two are WIP. Any comments/feedback/suggestions are welcome.

Wiki page is open in the player's default browser. The command respects the "Prompt on links" option from vanilla.

https://github.com/user-attachments/assets/1511870e-2b99-4de5-82b4-c2bc937d6cbf


Alexdoru:
also you need to think about adding tabcompletion options so that it's easier to use and we can see everything that's available

AbdielKavash:
> Looks like a nice addition to the wiki!

One of the goals is to encourage a positive feedback loop: players will have easier access to the wiki, they will see articles that are missing or incomplete or outdated, complain about them on discord, which gives us a chance to fix the wiki. Everyone wins!

Steelux8:
The more options you add for the command, the higher likelihood of the player stumbling upon more and more pages that don't exist. Having the options is nice, but I'm still not sure about the expectation of /wiki hand or /wiki block returning an empty page most of the time.

PR #1026, opened 2024-10-10
Hide render blocks from NEI

serenibyss:


Dream-Master:
@serenibyss you need get the itemems via itemlist (GT, Good Generator)

Dream-Master (approved):

PR #1030, opened 2024-10-11
Adds lapotron dust batch recipe and enderium skip

BlueHero233:
Adds a way to skip enderium base and a batch recipe for lapotron dust

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/MixerRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/MixerRecipes.java
@@ -490,20 +490,6 @@ public class MixerRecipes implements Runnable {
                 .itemOutputs(GTOreDictUnificator.get(OrePrefixes.dust, Materials.Enderium, 4L)).duration(10 * SECONDS)
                 .eut(8).addTo(mixerRecipes);
 
-        GTValues.RA.stdBuilder()
-                .itemInputs(
-                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Tin, 64L),
-                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Silver, 32L),
-                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Platinum, 32L),
-                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Thaumium, 64L),
-                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.EnderPearl, 64L),
-                        GTUtility.getIntegratedCircuit(12))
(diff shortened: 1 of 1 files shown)


chochem:
why a batch recipe? that is done on the level of machines

BlueHero233:
> why a batch recipe? that is done on the level of machines

went right over my head, let me fix that

BlueHero233:
> I don't think a Lapotron Dust skip is a good idea. I'm fine with Enderium though (but, why MV?)

Just a placeholder voltage, Enderium is LV though, maybe jot that up to HV and lower the time?

chochem:
I also changed the label on both because direct recipes are not an _enhancement_, but just a way to make the game easier. Hence balance.

src/main/java/com/dreammaster/gthandler/recipes/MixerRecipes.java
```diff
@@ -490,6 +490,20 @@ public void run() {
...
+                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Silver, 32L),
+                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Platinum, 32L),
+                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Thaumium, 64L),
+                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.EnderPearl, 64L),
+                        GTUtility.getIntegratedCircuit(12))
+                .itemOutputs(GTOreDictUnificator.get(OrePrefixes.dust, Materials.Enderium, 64L),
+                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Enderium, 64L),
+                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Enderium, 64L),
+                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Enderium, 64L)).duration(16 * MINUTES)
+                .eut(TierEU.RECIPE_MV).noOptimize().addTo(mixerRecipes);
```
Dream-Master, line 505: no sorry but not mv. To skip enderium bas it must be a few tier higher

Dream-Master (changes requested):

src/main/java/com/dreammaster/gthandler/recipes/MixerRecipes.java
Dream-Master: why github show this is outdated? its not

BlueHero233: Im in the middle of fixing it right now, probably gonna close this PR as the enderium recipe will be added in the ABS and the lapotron recipe removed

PR #1033, opened 2024-10-12
Move Digital Singularities from Dire table to Space Asssembler

StaffiX:
Moves the digital singularity recipe to space assembler to reduce the strain from dire tables and to make it more consistent to Artificial Universe.
image
image
image


serenibyss (approved):

EnderProyects:
Since you are with cells, could you check this? :3
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/17266

PR #1034, opened 2024-10-12
Move Solar Panels from Dire Table to Assembler/Assembly Line

IncandescenceSy:
Needing large quantities of Dire Table recipes has always been problematic in lategame, as Dire Autocrafting Tables are laggy and don't scale. Therefore, some of the most spammed recipes should be moved to more scalable machines, and Solar Panels are one of the worst offenders.

LV-IV panels are in the Assembler, and LuV-UV panels are in the Assembly Line due to running out of space. The recipe ingredients are unchanged, aside from the Assembly Line recipes including a small amount of Mutated Living Solder, as I believe all Assembly Line recipes need to have fluids. The research items are simply the previous tier of solar panel.

The recipes are 10 seconds and one voltage tier above the panel; I'd be open to changing this, as it may be too long for earlygame, especially considering they were previously effectively one tick with no power usage.

I've marked this PR as a draft because I want to get feedback on the possibility of changing the recipes more drastically; having 2-4 each of 5 different kinds of wafers/chips may make sense for a Dire Table recipe, but it feels a bit out of place in a GT machine. I'd like to redo the recipes to feel more sensible, without meaningfully changing the input cost, but I thought that that might be controversial, so I'd like to hear opinions on it. If people are in favor, I'll do it; if not, I can simply undraft the PR as is

Examples of the recipes:
image
image
image

### DIFF
--- a/src/main/java/com/dreammaster/scripts/ScriptGregtech.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptGregtech.java
@@ -29,16 +29,22 @@ import static gregtech.api.recipe.RecipeMaps.formingPressRecipes;
 import static gregtech.api.recipe.RecipeMaps.hammerRecipes;
 import static gregtech.api.recipe.RecipeMaps.maceratorRecipes;
 import static gregtech.api.util.GTModHandler.getModItem;
+import static gregtech.api.util.GTRecipeBuilder.HOURS;
 import static gregtech.api.util.GTRecipeBuilder.MINUTES;
 import static gregtech.api.util.GTRecipeBuilder.SECONDS;
 import static gregtech.api.util.GTRecipeBuilder.TICKS;
+import static gregtech.api.util.GTRecipeConstants.AssemblyLine;
+import static gregtech.api.util.GTRecipeConstants.RESEARCH_ITEM;
+import static gregtech.api.util.GTRecipeConstants.RESEARCH_TIME;
 import static gregtech.api.util.GTRecipeConstants.UniversalChemical;
 
 import java.util.Arrays;
 import java.util.List;
 
 import net.minecraft.item.ItemStack;
+import net.minecraftforge.fluids.Fluid;
 import net.minecraftforge.fluids.FluidRegistry;
+import net.minecraftforge.fluids.FluidStack;
 
 import com.dreammaster.gthandler.CustomItemList;
 
@@ -48,6 +54,7 @@ import gregtech.api.enums.GTValues;
 import gregtech.api.enums.ItemList;
 import gregtech.api.enums.Materials;
 import gregtech.api.enums.OrePrefixes;
+import gregtech.api.enums.TierEU;
 import gregtech.api.recipe.RecipeCategories;
 import gregtech.api.util.GTModHandler;
 import gregtech.api.util.GTOreDictUnificator;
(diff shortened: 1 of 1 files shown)


EnderProyects:
Why removing all the recipes of the dire table? Also, there is a dev that was working on a multi ( Reign)

IncandescenceSy:
Dire Tables are bad in every way possible, so moving spammed recipes out of it has become a goal of multiple devs lately. As far as I know, the multi hasn't gone anywhere, and I and many others agree that a dedicated multi with no unique mechanics (which is what Reign was planning) isnt interesting or the solution when machines such as the assembler already exist.

Keeping Dire Table recipes alongside GT machine recipes is just silly and adds pointless bloat and could confuse players (especially if the recipes end up getting revamped to fit the assembler better like I want)

EnderProyects:
Is just that like a month ago I read that four was planing on doing alternate recipes for solars in a new specializad solar multi, or something like that

IncandescenceSy:
Can you link me to when she said this?

EnderProyects:
Around this conversation
https://discord.com/channels/181078474394566657/181078474394566657/1288255403917840538

FourIsTheNumber:
> Is just that like a month ago I read that four was planing on doing alternate recipes for solars in a new specializad solar multi, or something like that
That conversation is me saying I am in favor of moving the slow recipes, generally. I'm not working on a dire/solar multi and if someone did we could migrate recipes into it later.

EnderProyects:
> That conversation is me saying I am in favor of moving the slow recipes, generally. I'm not working on a dire/solar multi and if someone did we could migrate recipes into it later.

I remember reading something about the old pr of the solar multiblock, so I just do 1+1 xd

IncandescenceSy:
It was PureBluez that made the Solar Factory quite a long time ago, but it got rejected immediately for numerous reasons. I don't know if he'll ever try to PR it again, and I have my doubts on the chances of it getting in if he does. A multi dedicated specifically to solar panels of all things seems odd and extremely unneeded, so I don't know what would justify its place. But on the off chance that it does happen, these recipes could just be moved or removed.

chochem:
the latest plan was not a solar multi (and not this) but a multi for direcrafting recipes.

chochem:
did you even realize you are changing the balance of the LV solar quite drastically?

but yea I dont think this should happen at all

IncandescenceSy:
> a multi for direcrafting recipes.

I haven't heard much about that lately, and it seems like many devs aren't on board with it anyway, so I'm not sure if that will actually happen. It also doesn't feel like it really makes sense; thematically what is the multi doing? It's just another stuff in -> stuff out magic box, but this time with a seemingly random selection of recipes with no explanation for what it's doing. It would certainly be an improvement over the current situation though, so I'll welcome it if it does happen, even if it's not what I'd want.

> did you even realize you are changing the balance of the LV solar quite drastically?

Surely you don't mean to suggest that you think someone would actually use LV solars for power just because it's craftable in MV? Oh no, it went from unusably terrible to still unusably terrible!

> but yea I dont think this should happen at all

Why not? Is it purely because you'd rather have a dedicated magic box for dire crafting?

> Maybe the endgame multi blocks just should not have the solars in them at all.

Unfortunately it's probably too late to change that...you'd have to rebalance a lot of things. That applies to many of the things that make the Dire Table problematic.

boubou19:
you shift the balance of crafting too much while doing so, and it's not the proper solution to fix this issue. Recursion should be broken into direct crafting if anything. And that would fix the issue only for the solar panels, which is a ductape. The main reason we all hate the Dire Autocrafting Table is because no matter how fast we can fill up its buffer, it's slow to output items. A simple change in this behavior and it would alleviate all the issues we have with it except recursion for crafts like solar panels. This PR would move away from the DAT, and i don't like it, as it remove automation diversity for yet another asslines / assemblers recipes.

StaffiX:
The slow to output issue is largely fixed by https://github.com/GTNewHorizons/Avaritiaddons/pull/15 so only unstackable items remains an issue after that.

IncandescenceSy:
If that PR gets merged, that will serve as enough of a band-aid fix for now until a better solution is agreed upon, whether it be a dire crafting multiblock, a dedicated solar multiblock, or perhaps 1-step recipes gated to UEV or so and using more advanced materials

PR #1035, opened 2024-10-12
Add a recipe to the Architect's Glowbrush

Superfrogman98:
Adds a recipe to the Glowbrush that was implemented in https://github.com/GTNewHorizons/ArchitectureCraft/pull/19


IncandescenceSy:
Why is there a gear in the recipe? That seems out of place to me

Superfrogman98:
I took the basis for the recipe from the buildcraft and extrautilities brushes which used a gear

IncandescenceSy:
My bad, I didn't know that the BuildCraft brush already used a Gear. It still doesn't really make sense, so I would have personally used something else, but there's precedent for it being included, so there's no issue

BlueHero233 (approved):
looks good 
image

Dream-Master:
Fine from my side to add the recipe since the other stuff allready in the pack.

PR #1036, opened 2024-10-12
Add recipes for Item Holder Covers

serenibyss:
Adds recipes for the Item Holder covers added here: https://github.com/GTNewHorizons/GT5-Unofficial/pull/2593.

Follows the format:
`Tiered Chest (from Input/Output Bus recipe) + Tiered Plate + C1 + Tiered Fluid (from other cover recipes), 40s @ 16 EU/t (both from other cover recipes)`

I decided not to have the assembler recipes be tiered, and instead all LV, but I could go either way on this so I'm open to opinions.

I'm marking this as a bugfix PR since these covers were added in this dev cycle, and recipes were just forgotten about for a while, so I think this should definitely go into 2.7.


Dream-Master:
It’s  fine add them early (lv is ok)

Dream-Master (approved):

PR #1037, opened 2024-10-12
readd iron ore gravel recipe

chochem:
was accidentally removed when ic2 recipes were moved.

also doesn't autogen for some reason and never has but the debug for that was accidentally(?) removed as well, so who knows.

also of course the block shouldnt even generate but thats not a bug anyone has figured out so far. XD

fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/17610


serenibyss (approved):

chochem:
> also doesn't autogen for some reason

oh right figured that part out. tconstruct is still excluded in the oredictevent https://github.com/nman1300/GT5-Unofficial/blob/4c80a553abaee3b8e3d0172c2bf04daad5a50e83/src/main/java/gregtech/common/GT_Proxy.java#L1739

PR #1039, opened 2024-10-13
Fix good circuit recipe

HoleFish:
The ingredient circuit was accidentally oredicted during cleanup


Dream-Master:
Why this should be not oredicted ?

FourIsTheNumber:
> Why this should be not oredicted ?

It's part of a circuit line and should only use the specific preceding circuit, not any LV. the output is an MV circuit.

chochem:
There is no bug in any refactor. The ic2 circuit was the unification in the past before I switched it to the lv-any.
So -for technical reasons- this was always oredicted.

Dream-Master:
@chochem if it should be ordicted the pr can be deleted ?

chochem:
I am happy with it. brings it in line with the other circuit lines. unless you dont want it of course.

Dream-Master (approved):

PR #1041, opened 2024-10-13
Add gtnh logo with wiki button and bug report button in the pause menu

Alexdoru:
image


src/main/java/com/dreammaster/mixin/mixins/early/MixinGuiIngameMenu.java
```diff
@@ -0,0 +1,86 @@
...
+        if (button.id == -161518) {
+            gtnh$openUrl("https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues");
+        } else if (button.id == -8998561) {
+            gtnh$openUrl("https://wiki.gtnewhorizons.com/wiki/Main_Page");
+        }
+    }
+
+    @Unique
+    private static void gtnh$openUrl(String address) {
+        try {
```
mitchej123, line 76: I'm not convinced this will work on anything other than windows.

I'd suggest looking at https://github.com/GTNewHorizons/Angelica/blob/master/src/main/java/net/coderbot/iris/gui/screen/ShaderPackScreen.java#L409 and making that available in NHlib as a cross platform util

OneEyeMaker: This way MC itself opens links. This piece of code is just copy of `GuiChat:func_146407_a`.

glowredman (changes requested):
While I think the idea is good, I don't like the placement. It's asymmetric without embracing it fully. In my opinion, putting these two in a corner would look better and reduce the risk of accidentally clicking on them.
The logo looks totally out of place too. It should either be in the middle between "Game menu" and "Back to Game" or not there at all (I'd prefer not there at all).
Finally, please use an event handler instead of mixins. Mixins may stop working at any moment, e.g. because someone decided it's a good idea to `@Overwrite` one or more of the methods you inject into. Hopefully this doesn't happen but maybe there'll be a good reason for it. We can't predict the future so it's best to use the intended means.

src/main/java/com/dreammaster/mixin/mixins/early/MixinGuiIngameMenu.java
mitchej123: I don't believe it works on Linux and they changed it to be more in line with what we have in Angelica on modern mc.  If it worked as is I wouldn't have gone to the effort to reimplement

glowredman (approved):

PR #1042, opened 2024-10-14
Add Cosmic Neutronium Block Solidifier Recipe

lordIcocain:
image


EnderProyects:
I think black hole compressor using the fluid suits better to neutronium, maybe using a stack recipe? That make 64 or more at the time?

lordIcocain:
why will use that if we have 
image

EnderProyects:
Because is different fluid than ingots, also does the EOH give fluid neutronium? ( I don't remember that's why I think they are different in some way)

FourIsTheNumber:
Fluid -> Block is really not suited to black hole, no. It doesn't do that kind of recipe and this would be wildly off-tier

lordIcocain:
> Because is different fluid than ingots, also does the EOH give fluid neutronium? ( I don't remember that's why I think they are different in some way)

but you can mold ingot then compressor. EOH didnt give, you get neutronium`s from DTPF

src/main/java/com/dreammaster/gthandler/recipes/FluidSolidifierRecipes.java
```diff
@@ -27,8 +28,12 @@
 
 public class FluidSolidifierRecipes implements Runnable {
 
+    ItemStack missing = new ItemStack(Blocks.fire);
```
chochem, line 31: why?

lordIcocain: As example i used CompressorRecipes.java. I repeated how it was done there.

chochem: 1) you are not using it 2) never use this for gt machine recipes, it just hides broken stuff

lordIcocain: fair.

FourIsTheNumber (approved):

PR #1044, opened 2024-10-18
Downtier wireless data hatches to UEV

StaffiX:
Wireless data currently has the same gating wireless computation, which is gated behind eternal DTPF, primarily because of dyson, but data doesn't have the same issue.
This downtiers the wireless data hatches to UEV, to help the assembly line scaling, for example crafting the DTPF, which usually makes the players setup a dozen of assembly lines.

image
image


NotAPenguin0 (approved):

NotAPenguin0:
Just needs conflicts to be resolved

PR #1045, opened 2024-10-18
Fix AL recipes for dyson parts, stargate, etc.

chochem:
they were only defined for specific circuits instead of allowing replacement. The other recipes (here!) look fine, only these 18 were bugged. There are more in GT probably.

fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/17671


TheEpicGamer274 (approved):
looks fine to me

PR #1046, opened 2024-10-18
Set hardness of the Ztones glass to be the same as normal glass

Alexdoru:
Currently the Ztones glass is insta break which makes it annoying to work with since you will constantly break some of it accidentally. Since it is insta break it will also often create ghost blocks. This pr sets the hardness of the block to be the same as vanilla glass.

### DIFF
--- a/src/main/java/com/dreammaster/main/MainRegistry.java
+++ b/src/main/java/com/dreammaster/main/MainRegistry.java
@@ -10,6 +10,7 @@ import static gregtech.api.enums.Mods.Thaumcraft;
 import static gregtech.api.enums.Mods.TinkerConstruct;
 import static gregtech.api.enums.Mods.TwilightForest;
 import static gregtech.api.enums.Mods.Witchery;
+import static gregtech.api.enums.Mods.ZTones;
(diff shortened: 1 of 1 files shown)


OneEyeMaker (approved):

S4mpsa:
I hate this.

Ultraxite:
more than anything in the world i want this reverted, having instant breaking glass with punch is exceptionally useful.

Dream-Master:
@S4mpsa which hardness you suggest ?

boubou19:
with a vajra there is no diff. idk why it's an issue?

S4mpsa:
It means you now need to switch to a Vajra when you do a mistake. Just a minor annoyance, but I suppose it is subjective seeing as Alexdoru sees it the other way around.

Ultraxite:
@boubou19 @Dream-Master being able to instant break it with just fists was very convenient. 

vajra is annoying to use especially for endgame when theres a lot of lag because u tend to break more than you intend when u left click with vajra. tinted glass breaking with fists meant you can hold down left click and only break glass and no other blocks.

PR #1052, opened 2024-10-21
Add steam machine hull assembler recipes

Dioxop:
Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/16915


Dream-Master:
@Dioxop you must use the right item via itemlist not via Moditem

chochem (changes requested):
those recipes just look incorrect, not sure what happened here. for example why does a standard 8 plate recipe suddenly need a fluid?
Also the likelihood that these are causing conflicts is like 90%+.

Dioxop:
I added a fluid because the steel machine hull recipe would cause a conflict otherwise and the other machine hulls also use fluids

chochem:
it didnt work, they are still conflicting lol.

Dioxop:
@chochem  how? I don't see any recipe that even uses bronze or steel plates with refined glue

chochem:
so you are looking for the wrong thing. that explains why you didnt catch conflicts. Anyway these recipes need to be changed fundamentally anyway.

Dioxop:
lol, okay. do you have a recommendation? i would have used an integrated circuit (8) if the steel hull didn't conflict with the lv machine casing

LazyFleshWasTaken:
I'd reocmmend eitehr c6 or c16, both should have no conflict with bronze or steel plates.

chochem:
it sounds like you found a bug and instead of reporting or fixing you are working very awkwardy around it.
(doesnt explain the rest of your recipes)

chochem:
specifically that the steel hull isnt craftable due to a conflict

chochem:
actually maybe the bug is that it exists and is used. we are normally using the bricked wrought iron hull

chochem:
for that one sure. but still need the others fixed

PS: totally didnt delete posts where I said stupid stuff because I am blind

Dioxop:
Okay, I implemented the recommended changes

src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
```diff
@@ -97,6 +97,40 @@ public void run() {
...
 
+        // Bronze Hull
+        GTValues.RA.stdBuilder()
+                .itemInputs(
+                        GTOreDictUnificator.get(OrePrefixes.plate, Materials.Bronze, 8),
+                        GTUtility.getIntegratedCircuit(6))
+                .itemOutputs(ItemList.Hull_Bronze.get(1)).duration(2 * SECONDS + 10).eut(TierEU.RECIPE_LV)
+                .addTo(assemblerRecipes);
+
+        // Bricked Bronze Hull
```
Ethryan, line 108: This conflicts with the Bronze Hull above if you are making a lot of them, since the only thing differentiating them is the addition of Bricks in this Recipe. Having them both on circuit 6 shouldn’t be done.

Dioxop: What are the exact rules for a new recipe? Does it need to have at least two different items or a different circuit?

RecursivePineapple: If any combinations of items (regardless of their amount) in a recipe can make another recipe, that's a conflict. If there's a conflict the recipe has to be changed to fix it, though this usually means changing the circuit.

Ethryan: This still hasn't been fixed.

Dream-Master: > This still hasn't been fixed.

I send him a message on Discord @Dioxop

serenibyss:
@Dioxop Any updates?

Dream-Master:
@Dioxop i send you a message twice. Can you check your pr? thanks

PR #1058, opened 2024-10-25
add recipe for infinity water cell

lordIcocain:
image


Nockyx:
I think this should be a dire table recipe, since it is quite an interesting creative item that you would craft only once

PR #1059, opened 2024-10-26
Added Infinity Water Cell Recipe

Nockyx:
The Item was added but given no recipe, this pr adds a dire crafting table recipe for it.

image
image


MCTBL:
Grate idea! but can we switch the fluid packet to other things else?
This cell only need one or two for each player, player won't make a auto crafting it for a long time, and fluid packet can only send
by interface. That's kindda suck.
Or can we make something made by bunch of water, like `water-singularity`-ish, cause water is nearly free, so we can make this core will take a loooooong time to craft, and this is the cost of this cell.

Nockyx:
I like the packet being acquired via an interface because it is a different way of creating something, but I'm not against creating a new item as some kind of water singularity, I'll look into that.

Nockyx:
for now i believe the fluid packets are the best option
they are a decent small "challenge" to get the hands on and it adds a little to this recipe.

Dream-Master (approved):

PR #1067, opened 2024-11-04
Remove Ancient Blocks from Chisel

54M44R:
Related to GTNewHorizons/ThaumicBases/pull/42

Meant to serve as a balancing change to prevent converting Iron, Gold, and Diamond blocks for free stability


src/main/java/com/dreammaster/scripts/ScriptChisel.java
```diff
@@ -224,6 +227,65 @@ public void loadRecipes() {
         ChiselHelper.addVariationFromStack("CobblestoneBricks", getModItem(Chisel.ID, "cobblestone", 1, 13, missing));
         ChiselHelper.addVariationFromStack("CobblestoneBricks", getModItem(Chisel.ID, "cobblestone", 1, 14, missing));
         ChiselHelper.addVariationFromStack("CobblestoneBricks", getModItem(Chisel.ID, "cobblestone", 1, 15, missing));
+        ChiselHelper.addGroup("diamond_block");
+        ChiselHelper.addVariationFromStack("diamond_block", getModItem(Minecraft.ID, "diamond_block", 1, 0, missing));
+        ChiselHelper.addVariationFromStack("diamond_block", getModItem(Chisel.ID, "diamond_block", 1, 1, missing));
```
OneEyeMaker, line 232: Why isn't block with meta 0 included into this group? And why don't you use `for` loop to simplify such repeated code?
Same for iron and gold.

54M44R: Some Chisel Blocks such as Iron, Diamond, and Gold start at meta 1 instead of 0. Not sure why that is

As for using for loops, I'm not sure why I didn't think of that. I'll throw that in

54M44R: Replaced with loops. I have refrained from touching the rest of the Scripts file which is unrelated to this PR

OneEyeMaker (approved):

PR #1070, opened 2024-11-05
Changed Ichorcloth Recipe to remove screws and screwdriver

Nockyx:
The current ichorcloth recipe uses screws and a tool in the recipe, even tho this isn't exactly a problem it appears very weird and bothers some people to have to use the nc tool.

This pr changes the recipe to use 1 osmiridium foil instead, so keeping the OsIr cost, and adding 2 void foils to mimic the ichorium ingot recipe. The cost of void metal is irrelevant.

image


AbdielKavash:
Could you swap the positions of Ichor and Void Foils? That would make it resemble the recipe for Ichorium Ingots more:



Otherwise I like this change, tools in arcane crafting recipes mess with AE calculations. It's fine for items you only ever need one or a few of, but basic resources that one might need in bulk should have a tool-less recipe (or at least an alternative).

LazyFleshWasTaken:
> Otherwise I like this change, tools in arcane crafting recipes mess with AE calculations. It's fine for items you only ever need one or a few of, but basic resources that one might need in bulk should ideally have a tool-less recipe (or at least an alternative).

Agreed, though the single-use tools added definitely help too.

src/main/java/com/dreammaster/scripts/ScriptThaumicTinkerer.java
```diff
@@ -1706,17 +1706,17 @@ public void loadRecipes() {
...
                 getModItem(ThaumicTinkerer.ID, "kamiResource", 1, 0, missing),
                 'd',
-                "gemFlawlessDiamond",
+                "foilVoid",
                 'e',
-                getModItem(Thaumcraft.ID, "ItemResource", 1, 7, missing),
+                "foilOsmiridium",
                 'f',
-                "gemFlawlessDiamond",
+                "foilVoid",
```
Dream-Master, line 1713: why void foil is called twice.
letter d and f need only one entry
same for the others
(i see many recipes follow this logic but so far i know is this obsolete code)

Nockyx: i just followed the way it was alr defined. i can change to reflect more modern applications for sure.

Nockyx:
> Could you swap the positions of Ichor and Void Foils? That would make it resemble the recipe for Ichorium Ingots more:
> 
> 
> 
> Otherwise I like this change, tools in arcane crafting recipes mess with AE calculations. It's fine for items you only ever need one or a few of, but basic resources that one might need in bulk should ideally have a tool-less recipe (or at least an alternative).

for sure, i'll change it to mimic the ingot better

Nockyx:
image
new recipe

AbdielKavash (approved):

PR #1089, opened 2024-11-20
Fix golembell

chochem:
fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/17996

goes together with a PR in Thaumic Horizons as there are multiple layered bugs.


chochem:
testing a bit more, I looks like there is yet another layer to this as there is still one extra recipe generated somewhere. will fix in a bit

chochem:
alright, all ready.

Dream-Master (approved):

PR #1096, opened 2024-11-27
Add more recycling recipes

Ableytner:
closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/17293
Also adds a recycling recipe to Reinforced Stone from IC2.


chochem (dismissed):
recycling recipes should be added via itemdata. that means using `GT_OreDictUnificator.addItemData` or `GT_OreDictUnificator.addItemDataFromInputs`.

This gives the multiple recycling recipes, puts them in the right tabs, and allows inheritance of the data to other recipes.

see for example https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/936

chochem (approved):
would have been easier to let the code do the math but this is fine.

PR #1104, opened 2024-12-14
Fix Typos of Slime Sou(i)l

Yoshy2002:
hope i got everything
closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/17909


Ethryan (commented):
Hmm, I think this change will delete all already existing slime soul items, since it looks like the newly named item would be added as a new one.

Yoshy2002:
> Hmm, I think this change will delete all already existing slime soul items, since it looks like the newly named item would be added as a new one.

my bad, didnt think about that but they infact get deleted on world load
If there is no easy fix for that, ill close my PR.

PR #1106, opened 2024-12-14
Fix avaritia tool recipes in NEI

chochem:
Fixes a long standing bug where avaritia tools recipes dont show directly in NEI.
Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/13386 and a bunch of duplicates.

that particular NBT being set to 0 or not existing is fully functionally equivalent.


YannickMG (approved):
Works as advertised.

PR #1107, opened 2024-12-15
stop Luv solenoid being gated ZPM

Yoshy2002:
as described in the issue, 1080k Sp Coolant needs Fluxed Elextrum, which needs Trinium Coils which needs ZPM Assembler in LuV
changed the 1 1080k Sp to 2 540k Sp Coolant Cells
image
closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/18201


RecursivePineapple (approved):
Looks good, don't know how I missed that

PR #1109, opened 2024-12-16
Add recipes for preconfigured p2ps

EnderProyects:
fix https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/16837

image
image
image
image
image
image
image
image

### DIFF
--- a/src/main/java/com/dreammaster/scripts/ScriptAppliedEnergistics2.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptAppliedEnergistics2.java
@@ -3,12 +3,16 @@ package com.dreammaster.scripts;
 import static gregtech.api.enums.Mods.AppliedEnergistics2;
 import static gregtech.api.enums.Mods.Avaritia;
 import static gregtech.api.enums.Mods.BartWorks;
+import static gregtech.api.enums.Mods.BuildCraftFactory;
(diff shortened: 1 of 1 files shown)


EnderProyects:
> These recipes seem okay, though I wonder if there's a cleaner way to do it to avoid the added cost without needing to do something like 16 at a time.
> 
> What about allowing these P2Ps to chisel into eachother? Then you could also automate it by using multiple chisel buses as you see fit on a multiblock autochisel. (this is similar to modern ae2 as well, which does it with the stonecutter)

Well the other parts from ae2 that have variants are also crafted in circuit assembler, so I dont see why this should be different, also you are only needing this ones if you are doing things like assline or other stuff so mainly craft a lot, p2p are cheap, and the items used to change them are also very cheap

EnderProyects:
@serenibyss If you think that using this recipes is annoying or a nerf I could change them with chisel ( well I could but the attempts I have done with chisel haven't work, so I prefer not doing that) if not are you are okay with the recipes, I would like them to be in the next release ( that I think is tomorrow) 🙃

NotAPenguin0 (approved):
Really nice QoL

EnderProyects:
@boubou19 could this be added to  .4?

PR #1110, opened 2024-12-16
Add recipe for inf lava cell

EnderProyects:
Added a recipe, similar of the inf water one, but red, fiery and magic gated, since lava can not be "created", so it uses the magic of the Everburn Urn, powered by a Lava sigil, using crucibles of fire as magic conduicts ( lore) 
image

Needs this in order to work: https://github.com/GTNewHorizons/AE2FluidCraft-Rework/pull/250


CactusCool:
this just gives you infinite power with EHE.

serenibyss:
Closing, see https://github.com/GTNewHorizons/AE2FluidCraft-Rework/pull/250#pullrequestreview-2506970601

Pxx500:
just dig lava on ross, this would be metabreaking

EnderProyects:
> find something else than fluid packets pls, it's annoying to have to make a subnet with regular interface with fluid patterns just to get the items.

I just placed them because of the similarity with the inf water recipe, my idea if it was aproved was to use the Geologic and Hydrolic singularities as Lava and Water singularities and add the recipes for them, changing the recipes that uses packets

PR #1120, opened 2024-12-24
Add Recipe for Solar Factory

purebluez:
Related PR: https://github.com/GTNewHorizons/GT5-Unofficial/pull/3703

Of course, do not merge this unless the related PR is merged.
Also, please close if the related PR is closed.

The Recipe, which is also shown in the related PR.
image

The recipes Solar Factory do have been moved here. See the bottom of related PR description for what they look like.

Closes https://github.com/GTNewHorizons/areas-of-interest/issues/6


EnderProyects:
You could add a luv solar panel to the recipe so in order to craft the controller it needs some resources spended in the normal way ( this also can make new players see that this recipes are much better) 😜

purebluez:
Forgot to do this before
Converting to draft since related pr is also a draft

Dream-Master:
image
can you add this tem as well. ik we remove super solar mod some day but we well readd items to core mod

Dream-Master:
zpm missing asoc wafers 
image

Dream-Master:
I see recursive recipes use always tier 1 and advanced version always t3. Can you change this so Lv-hv use  ev-luv t2 and zpm to uv t3. What about t4 four? i see it in 3d nei

purebluez:
> What about t4 four? i see it in 3d nei

4th tier in the NEI structure viewer is just 3rd multi tier but with t4 precise casing. I would disable it showing the "4th tier" if I knew how, since there is no tier 4 in code.
The rest of your changes suggested I will implement. Thank you for the feedback!

Dream-Master (dismissed):
@purebluez why the buff? yea ik the recipes a bit more expensive but i not support this buff

purebluez:
> why the buff? yea ik the recipes a bit more expensive but i not support this buff

https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/18925 complained about the prices and after some thought i decided the recursive recipes are best not being buffed, so they are now 1:1 with dire table 

Im certain the wafer mechanic would not shift balance much for powergen, as you need to be at least a tier higher than what the recipe calls for. For example, if HV solar calls for Phosphorus wafers, to get an output multiplier you would need a Naquadah wafer which is tiered to LuV. Not saying the multiplier is useless either, it will let you save alot of material if you make solars for use in crafting recipes like Dyson parts.

For the skipping recipes, those are intended to be made 2 tiers higher than the normal recipe. You can see on the recipe table that MV solar for example uses some ingredients from the EV solar. The point of the skipping recipes is solely convenience, and id say by the time you can make an EV solar there's no point to running the skipping MV recipe for powergen, where the balance really matters.

Killermaschine88:
Opinion on skip recipes being the same as tiered recipe but required all circuits +1 tier?

purebluez:
> Opinion on skip recipes being the same as tiered recipe but required all circuits +1 tier?

Honestly, i am starting to think the skipping recipes should not exist. Biggest balance problems seem to be with them, and their utility kind of goes away if the panels can be crafted with a normal pattern. Plus it removes magic ingredients from certain panels, and I view putting them back in as a bit of a hack.

Killermaschine88:
> > Opinion on skip recipes being the same as tiered recipe but required all circuits +1 tier?
> 
> Honestly, i am starting to think the skipping recipes should not exist. Biggest balance problems seem to be with them, and their utility kind of goes away if the panels can be crafted with a normal pattern. Plus it removes magic ingredients from certain panels, and I view putting them back in as extra ingredients as a bit of a hack.

then should just remove the recipes from dire table entirely cause dire table is a wack its so limiting and just move dire recipes to solar factory

serenibyss (approved):

PR #1121, opened 2024-12-24
First round of EFR recipe integration

chochem:
removes a bunch of vanilla recipes and adds new ones for banner conversion, cherry wood trapdoor, and observer.

will be continued some other time but that will be other PRs


src/main/java/com/dreammaster/scripts/ScriptEFR.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public void loadRecipes() {
+        long bits = GTModHandler.RecipeBits.NOT_REMOVABLE | GTModHandler.RecipeBits.REVERSIBLE
+                | GTModHandler.RecipeBits.BUFFERED;
+        GTModHandler.addCraftingRecipe(
+                GTModHandler.getModItem(EtFuturumRequiem.ID, "observer", 1L),
+                bits,
+                new Object[] { "AEA", "BCD", "AAA", 'A', "cobblestone", 'B',
+                        GTModHandler.getModItem(Minecraft.ID, "comparator", 1, 0), 'C', ItemList.Sensor_LV, 'D',
```
serenibyss, line 36: LV Sensor for an observer seems too expensive

EnderProyects: if at least instead of using cobblestone it used compressed cobble and it gives 8 or 16 ...

chochem: we are not adding stuff like this with vanilla recipes and LV is hardly late. But I can have a look at our recipes for similar modded blocks and go more in that direction

chochem: changed the recipe

serenibyss (approved):
Recipes and code seem reasonable to me now, I'll leave this open for @Dream-Master to give the final say

Dream-Master:
let me find out what the observer block does

PR #1124, opened 2024-12-26
Project red tinker fix

chochem:
fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/18513

these changes are no longer relevant due to https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/18480, I did in fact do some testing there but missed this.


Ethryan (approved):
This fixes it, right?

PR #1129, opened 2024-12-29
Change Bio Vat Recipes to use new metadata

StaffiX:
Requires https://github.com/GTNewHorizons/GT5-Unofficial/pull/3728
Makes coremod bio vat recipes to use new metadata


serenibyss (approved):
Will need a new GT5u tag

PR #1149, opened 2025-01-11
Add Chromatic Glass Crystal

Nockyx:
This PR adds Chromatic Glass Gems; Normal, Flawless and Exquisite.
image
These will be used in the Infinity Line.

The Material System that Chromatic Glass is created with doesn't support gem generation inside GT5u and i couldn't get it to work there. So decided to add it in coremod, alongside the already existing lens. Just as it is done with Mysterious Crystal.


Dream-Master:
have this items any use or you planing something @Nockyx

Nockyx:
> have this items any use or you planing something @Nockyx

> These will be used in the Infinity Line.

These Crystals will be made and used within the Infinity rework that is planned https://discord.com/channels/181078474394566657/1307066387096666242

Ethryan:
Hmm, Maybe reduce the amount of coloring a tad, or fix the coloring of the gems, since the flawless gem looks a lot more like a lens.
Since the gem lines are almost hidden on that one.

serenibyss (approved):

PR #1150, opened 2025-01-11
com.dreammaster.item.ItemList -> NHItemList

YannickMG:
Rename com.dreammaster.item.ItemList to NHItemList in order to avoid having to spell out the fully qualified class name 529 times in Dreamcraft, because it's almost always used in conjunction with GT's ItemList.

In the few files it isn't used with GT's ItemList, it'll make it clear that it's the Dreamcraft version.

This will undoubtedly disrupt currently open PRs but it's an easy fix for them, and I think it's worthwhile.

### DIFF
--- a/src/main/java/com/dreammaster/scripts/ScriptNatura.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptNatura.java
@@ -12,7 +12,6 @@ import static gregtech.api.recipe.RecipeMaps.maceratorRecipes;
 import static gregtech.api.util.GTModHandler.getModItem;
 import static gregtech.api.util.GTRecipeBuilder.SECONDS;
 import static gregtech.api.util.GTRecipeBuilder.TICKS;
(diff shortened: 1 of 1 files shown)


serenibyss:
@YannickMG These two files will need to be updated as well:
- https://github.com/GTNewHorizons/nei-custom-diagram/blob/f38d94cdce4a5d5894054c94d74a01559d9db176/src/main/java/com/github/dcysteine/neicustomdiagram/generators/gregtech5/circuits/CircuitLineHandler.java#L136
- https://github.com/GTNewHorizons/Crops-plus-plus/blob/31cf8d27fa491a8e6e8d70476358f23914f015e0/src/main/java/com/github/bartimaeusnek/cropspp/GTHandler/GTNHMachineRecipeLoader.java#L566

YannickMG:
Good catch, I just assumed nothing depended on Dreamcraft.

chochem (approved):
maybe gdc/nockyx will have to adjust their PR, but worth it.

PR #1156, opened 2025-01-12
Add Circuit Indicating Catalyst Tier for DTPF Recipes

Vlamonster:
This change introduces circuit indicators for DTPF recipes, mapping catalyst tiers to specific circuit values.

Default Mappings:
- DTCC: 1
- DTPC: 2
- DTRC: 3
- DTEC: 4
- DTSC: 5

Combs:
- DTCC: 11
- DTPC: 12
- DTRC: 13
- DTEC: 14
- DTSC: 15

Cosmic Neutronium (subset of LuV Superconductor):
- DTCC: 21
- DTPC: 22
- DTRC: 23
- DTEC: 24

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/DTPFRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/DTPFRecipes.java
@@ -62,7 +62,7 @@ public class DTPFRecipes implements Runnable {
 
                 long tier_1_quantity = 144L * base_quantity;
 
-                GTValues.RA.stdBuilder()
+                GTValues.RA.stdBuilder().itemInputs(GTUtility.getIntegratedCircuit(1))
                         .fluidInputs(
                                 MaterialsUEVplus.ExcitedDTCC.getFluid(neutronium.getCatalystAmount(0)),
                                 Materials.Iron.getMolten(tier_1_quantity))
@@ -75,7 +75,7 @@ public class DTPFRecipes implements Runnable {
 
                 long tier_2_quantity = 144L * base_quantity * tier_up_multiplier;
 
-                GTValues.RA.stdBuilder()
+                GTValues.RA.stdBuilder().itemInputs(GTUtility.getIntegratedCircuit(2))
                         .fluidInputs(
                                 MaterialsUEVplus.ExcitedDTPC.getFluid(neutronium.getCatalystAmount(1)),
                                 Materials.Iron.getMolten(tier_2_quantity))
@@ -88,7 +88,7 @@ public class DTPFRecipes implements Runnable {
 
                 long tier_3_quantity = 144L * base_quantity * tier_up_multiplier * tier_up_multiplier;
 
-                GTValues.RA.stdBuilder()
+                GTValues.RA.stdBuilder().itemInputs(GTUtility.getIntegratedCircuit(3))
                         .fluidInputs(
                                 MaterialsUEVplus.ExcitedDTRC.getFluid(neutronium.getCatalystAmount(2)),
                                 Materials.Iron.getMolten(tier_3_quantity))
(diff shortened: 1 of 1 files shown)


Vlamonster:
Comes together with [#3791](https://github.com/GTNewHorizons/GT5-Unofficial/pull/3791).

Vlamonster:
I forgot to put the circuit last for some recipes. Will fix tomorrow unless someone else does it.

joshaup1999:
What is the reason for adding an extra 14 circuits to the DTPF? Won’t think increase recipe checks on the multi?

Ruling-0:
What would these even do? Given a significant portion of the recipemap is pure fluids, so this couldn't improve stocking on-demand setups. Single catalyst stocking setups would be totally unchanged, and this would just make crib setups need to use multiple cribs. I don't see the benefit.

Vlamonster:
Makes it so specifying the circuit uniquely determines which catalyst is used, and helps with stocking fluids for convergence without conflicts. Additionally, you wouldn't really need more cribs, because you'll most likely only be using the highest tiers catalyst. Furthermore, for simplicity players could use the following 3 crib setups to achieve all recipes {C01, C02, C03, C04, C05}, {C11, C12, C13, C14, C15}, {C21, C22, C23, C24}. 

I don't know what stocking on-demand is though. It seems a bit oxymoronic 😅 .

Ruling-0:
> Makes it so specifying the circuit uniquely determines which catalyst is used, and helps with stocking fluids for convergence without conflicts. Additionally, you wouldn't really need more cribs, because you'll most likely only be using the highest tiers catalyst. Furthermore, for simplicity players could use the following 3 crib setups to achieve all recipes {C01, C02, C03, C04, C05}, {C11, C12, C13, C14, C15}, {C21, C22, C23, C24}.
> 
> I don't know what stocking on-demand is though. It seems a bit oxymoronic 😅 .

Convergence can be specified already in the UI or by only using the catalyst for the recipe you want to use.

That's still a few more cribs than 1.

Stocking on demand I meant as an on-demand setup that uses adv. stocking.

On further thought, there is a benefit. Stocking on-demand setups can now stock all catalysts, and using circuit prio, guarantee that recipes that can use stellar use it while recipes that can only use exotic use that, without needing toggle busses. Whether this is worth a change I leave to the admins. What might be alternatively nice is adding stellar to a few more recipes (namely dark matter).

Vlamonster:
> > Makes it so specifying the circuit uniquely determines which catalyst is used, and helps with stocking fluids for convergence without conflicts. Additionally, you wouldn't really need more cribs, because you'll most likely only be using the highest tiers catalyst. Furthermore, for simplicity players could use the following 3 crib setups to achieve all recipes {C01, C02, C03, C04, C05}, {C11, C12, C13, C14, C15}, {C21, C22, C23, C24}.
> > I don't know what stocking on-demand is though. It seems a bit oxymoronic 😅 .
> 
> Convergence can be specified already in the UI or by only using the catalyst for the recipe you want to use.
> 
> That's still a few more cribs than 1.
> 
> Stocking on demand I meant as an on-demand setup that uses adv. stocking.
> 
> On further thought, there is a benefit. Stocking on-demand setups can now stock all catalysts, and using circuit prio, guarantee that recipes that can use stellar use it while recipes that can only use exotic use that, without needing toggle busses. Whether this is worth a change I leave to the admins. What might be alternatively nice is adding stellar to a few more recipes (namely dark matter).

The old recipes still required a minimum of 3 cribs to have no conflicts, so that hasn't changed.

serenibyss:
I don't think this is a good idea, I see little meaningful benefit from this, but it would be very disruptive to current setups and make them messier to fix as well. These sorts of wide programmed circuit maps like this are something we should be trying to reduce not increase

Vlamonster:
It should be very easy to migrate. On our private server it required no changes. The benefits would be to allow for more control over which catalyst is used - very useful since not all recipes currently take DTSC leaving you with more than one catalyst to choose from in the end-game. CRIBs will see less benefit, but setups using advanced stocking hatches will see great benefit.

serenibyss:
I would rather see the recipes that don't take stellar get stellar recipes than do a change like this PR

Vlamonster:
That would still make some automation annoying - e.g. infinity is a lot quicker with lower tier catalysts until a lot later on.

boubou19:
This brings no real benefits other than breaking current recipes on update. I'm closing this.

PR #1159, opened 2025-01-12
Streamline smeltery melting recipes

YannickMG:
I'm going to be working in those files and the huge repetition annoyed me. Trying to lessen it somewhat.

Incidental changes: 
- redGranite cassiterite melted into half the amount as other versions of this ore for some reason, made it the same.
- For some reason rubber melted into glue at different temperatures. That didn't seem meaningful and worth keeping the extra calls so I unified it.

There should be no other functional changes. This should make my life easier when changing aluminium brass to brass as a casting material.


serenibyss (approved):
Nice find on those two inconsistencies

PR #1167, opened 2025-01-16
New ae2 content recipes

serenibyss:
|              **Item**              |                                         **Recipe**                                        |                                                           **Explanation**                                                          |
|:----------------------------------:|:-----------------------------------------------------------------------------------------:|:----------------------------------------------------------------------------------------------------------------------------------:|
| **Creative Energy Anchor**         | image | _Does what it sounds like, so this recipe isn't complicated_                                                                       |
| **Fake Crafting Card**             | image | _Follows basic pattern of other card recipes. Uses a GT crafting table cover for extra flavor over a Crafting Table or similar_    |
| **Superliminal Acceleration Card** | image | _Follows the recipe of the Hyper-Acceleration Card, but using a UV World Accelerator and Dilithium Gem instead to gate it further_ |
| **Overflow Void Card**             | image | _This is effectively a void cell in a card, so the recipe represents that_                                                         |
| **Equal Distribution Card**        | image | _Uses an EV Item Distributor from GT in the recipe, since the name fits and it does a similar behavior as this new card_           |

### DIFF
--- a/src/main/java/com/dreammaster/scripts/ScriptAppliedEnergistics2.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptAppliedEnergistics2.java
@@ -1853,6 +1853,40 @@ public class ScriptAppliedEnergistics2 implements IScriptLoader {
                 "plateNetherQuartz",
                 "gemNetherQuartz",
                 "craftingToolFile");
+        // Creative Energy Anchor
+        addShapedRecipe(
+                getModItem(AppliedEnergistics2.ID, "item.ItemMultiPart", 1, 690, missing),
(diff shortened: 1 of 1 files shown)


EnderProyects:
In the fake crafting card instead of using a normal blank card why not using a crafting card, so now we can use them for something at least, same for the speed one, it could use a speed card

serenibyss:
Also added a recipe for wireless hub:
8 Wireless Connectors + 1 LuV Emitter -> 1 Wireless Hub (10s @ IV in assembler)

FourIsTheNumber (approved):

PR #1172, opened 2025-01-20
Recipe for quantum ring card

EnderProyects:
Add recipe for quantum ring card
https://github.com/GTNewHorizons/AE2FluidCraft-Rework/pull/210
image

Players will now be able to get wireless terminal without inf card, still it uses quantum rings and need one for each players, so is not as powerfull as inf card

Change inf card to use QR cards

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,7 +1,7 @@
 // Add your dependencies here
 
 dependencies {
-    api("com.github.GTNewHorizons:GT5-Unofficial:5.09.51.83:dev")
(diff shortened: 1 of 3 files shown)


NotAPenguin0:
Why can't players just build a quantum ring and put a wireless access point on it instead of adding another weird card?

EnderProyects:
The card was added long ago, it just didnt have a recipe

EnderProyects:
Btw closing this because somebody forgot to mention that while the name and description of the card says it conects to a quantum ring, it does not that, and just works the same as the inf card

PR #1173, opened 2025-01-21
Cursed recipe to get milk

EnderProyects:
Add one ae2fc recipe to turn drops of automagy milk into normal milk, is the only way I got to "register" the automagy fluid milk

Solves this
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/16694#issuecomment-2603495975

I know is very very cursed, but it works!! And well since is for a magic mod you can say that the milk got warp :D


serenibyss:
This is not an acceptable way to fix this issue, recipes using fluid drops like this is not a step in the right direction even if it does technically fix a minor issue. Not to mention that this doesn't help in any way players that make a thirsty tank for milk for their backpack in earlier game

PR #1174, opened 2025-01-21
Fix loop between Optical Processor and Tier 3.5 Memory

EnderProyects:
Added another recipe for the memory that uses the "old" recipe materials without the circuit

Solves this https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/18832

image

Also changed the numbers in the code, it gives 32


serenibyss (changes requested):
Instead of adding an alternate recipe, lets just change the current recipe from 1 UV circuit to 2 ZPM circuits.

EnderProyects:
Changing the current one will break current setups, thats why I added a second one, but okay

Dream-Master:
First of all i want to know why this recipe need a change?

serenibyss:
> First of all i want to know why this recipe need a change?

This recipe was originally added for optical circuits since tier 3.5 memory is used for optical memory. However the first optical circuit is UV, so if you use only optical circuits for UV-UIV tier circuits, then you need optical circuits to make optical circuits because of this recipe

src/main/java/com/dreammaster/gthandler/recipes/SpaceAssemblerRecipes.java
```diff
@@ -293,9 +293,12 @@ public void run() {
...
                                 ItemList.Circuit_Chip_SoC.get(64),
                                 ItemList.Circuit_Chip_NAND.get(64),
-                                GTOreDictUnificator.get(OrePrefixes.circuit, Materials.UV, 1L),
+                                GTOreDictUnificator.get(OrePrefixes.circuit, Materials.ZPM, 2L),
                                 GTOreDictUnificator.get(OrePrefixes.foil, Materials.VanadiumGallium, 64L))
-                        .fluidInputs(new FluidStack(solderIndalloy, 1152))
+                        .fluidInputs(
+                                new FluidStack(solderIndalloy, 1152),
+                                new FluidStack(solderUEV, 720),
+                                Materials.Gold.getMolten(16 * 144L))
```
serenibyss, line 301: Please revert this portion of the change

serenibyss (approved):

PR #1178, opened 2025-01-22
Magnet Dislocator recipe

EnderProyects:
Add dislocator recipe
Depends on this one
https://github.com/GTNewHorizons/Draconic-Evolution/pull/60

image

Changed obsidian slabs into plates, since I read somewhere that is planed to remove railcraft ( obsidian slabs are from railcraft)


Dream-Master:
WHO wrote this ?

EnderProyects:
I remember some months ago reading in general chat some devs saying that railcraft was going to be reomved in the future from the pack, I can change it back to slabs, is kinda the same cost of obsidian

Dream-Master:
> I remember some months ago reading in general chat some devs saying that railcraft was going to be reomved in the future from the pack, I can change it back to slabs, is kinda the same cost of obsidian

No why. Its not what i planing. Or i am not involved in it but without my permission i dont remove any mod.

mitchej123:
Seems like random banter
image

serenibyss (approved):

PR #1181, opened 2025-01-24
Replace Aluminum Gravel Ore with Zinc Gravel Ore, and deal with the consequences

YannickMG:
Fulfills **Area of Interest: Replace ALuminium gravel with another material to eliminate tier problems in GT.** Read https://discord.com/channels/181078474394566657/1323615140880580608 for the full story.

Some of the changes shown here are implemented in [the modpack PR](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/18858) and in [the GT5 PR](https://github.com/GTNewHorizons/GT5-Unofficial/pull/3852). They should be merged together.

### Zinc Gravel Ore
image
image
Replaces Aluminum in XP bucket recipes.
image

### Brass replaces Aluminum Brass almost everywhere
image
image

### Except the Aluminum Brass Singularity, replaced by a Rose Gold Singularity
rose-gold-recipe
meteoric-singularity-recipe

### Postea converstion
Handles Aluminum Gravel Ore, solid Alumnium Brass items and blocks and Molten Aluminum Brass in smelteries, nothing else.

#### Before
before

#### After
after

### Alumite is renamed to Obzinite and has it's recipe changed.
Zinc replaces Aluminium in equal quantities. TiCo parts with Alumite are unaffected and need no conversions.
image
image
image

Searching for Alumite still lets you find Obzinite for a seamless transition.
image

### Tinker manuals are updated
image

### Cobalt Brass uses Zinc instead of Aluminum
 I used tin instead of Zinc because Zinc is already present in brass, and I've seen that Tin can be an IRL brass component.
image
image
image

### Out of scope for this set of PRs.
- Quests will be done after everything is merged. (and not by me)
- Aluminium ore berries and raw alumimium are untouched, but will be made available later like in MV or at the earliest in the twilight forest through quest updates.
- Magnalium can have it's quest moved to the Twilight Forest.

### Final word
I'll think twice before starting to work on such a "simple" change again.


src/main/java/com/dreammaster/tinkersConstruct/TiCoLoader.java
```diff
@@ -1,13 +1,76 @@
...
 public class TiCoLoader {
 
+    public static void overrideAlumiteMaterial() {
+        Materials.Alumite.mDefaultLocalName = "Obzinite";
+        Materials.Alumite.mMaterialList.clear();
+        Materials.Alumite.mMaterialList.addAll(
+                ImmutableList
+                        .of(new MaterialStack(Zinc, 5), new MaterialStack(Steel, 2), new MaterialStack(Obsidian, 2)));
+        Materials.Alumite.mChemicalFormula = Materials.Alumite.mMaterialList.stream().map(MaterialStack::toString)
+                .collect(Collectors.joining()).replaceAll("_", "-");
```
serenibyss, line 48: We should make this change in 5u rather than here, that way we can also rename the variable and make things more clear

serenibyss (commented):
Initial brief review, I'll go through this in-depth when I have some more time this weekend

src/main/java/com/dreammaster/tinkersConstruct/TiCoLoader.java
YannickMG: Since GT5 is GTNH-only, which I forgot, it is the right place to make those changes so I will move them there.

That being said I would rather not change either the variable or the material name for a few reasons:
- Keeping the material name so that it unifies with TiCo Alumite.
- Keeping the material name so that searching for "Alumite", which most players will do, still works.
- Keeping the variable name to help in future maintenance, since other Tinker's based mods will still refer to it as Alumite and we can't change the variable name in those  since they're not GTNH-only. This material is still Tinker's Alumite, but it gets renamed in display text for GTNH reasons.

serenibyss (approved):
Code looks good, I tested this and everything seems to work well. Nightly testing seems to have gone fine as well

PR #1179, opened 2025-01-23
Asgardandelion 2.0

EnderProyects:
Since there are masses of angry people in general chat, I changed the flowers to runes, I also decreased the number of infinity plates since tablet and pool uses 6, and previously this used 12, now it uses 4 + 8 nuggets, also I added some more rainbow to the recipe

PR #1182, opened 2025-01-24
Asgardandelion 3.0 or whatever

EnderProyects:
- Changing to addRecipe() instead of addExtremeShapedOreRecipe()
- Decrease inf plates cost to be more inline with mana pool and tablet
- Closing this I guess https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/1179
image


serenibyss:
We decided to take #1179 instead, so closing this PR

PR #1186, opened 2025-01-31
A shit recipe for animal waste

EnderProyects:
image

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,11 +1,11 @@
 // Add your dependencies here
 
 dependencies {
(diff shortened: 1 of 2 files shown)


Dream-Master:
@EnderProyects why this is linked with the draft pr ?

NotAPenguin0:
What? Why would you need massive amounts of this using anything other than the tech around the intended tier of the line (LV/MV)

serenibyss:
Fully agree, I see no reason to add this recipe when all it does is bypass the intended way to make it

PR #1190, opened 2025-02-06
Add botania into inf gear recipes

EnderProyects:
Basically this:

closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/18856


combusterf (changes requested):
The description is disingenuous: a wide variety of items are introduced, plenty of which are not from Botania.

Also a reminder that Botania is a requirement for draconic evolution, and its technologic completion is thus already enforced.

Ruling-0:
It would be nice to include before and after images of the recipes so it's more obvious what was changed.

Dream-Master:
@EnderProyects yes add some old and new recipe pictures and we can start to discuss.

combusterf:
@EnderProyects The real problem is that you have made a reputation for yourself, and less about this being a beneficial change or not. I'm expecting you to be honest upfront and not do something that looks like you're trying to sneak in some extra nerfs on the side.

So do what we expect every good change request to look like: post the exact changes, and mention how you think it affects balance (including nerf, buff, etc.) Once that's done and checks out, I'll unblock my review, and if you're lucky, approve it if you managed to convince me.

Ruling-0:
Primordial leggings being the only piece of the set to show up is weird. What about instead putting the terrasteel leggings where the sigil is, and moving the sigil up one (replacing a spnt plate)?

EnderProyects:
> Primordial leggings being the only piece of the set to show up is weird. What about instead putting the terrasteel leggings where the sigil is, and moving the sigil up one (replacing a spnt plate)?

ok ;3

Ruling-0:
I've been thinking on this for a bit. I don't mind the gear changes but I don't like the wand update, personally. I think the recipe is just a bit uglier (misses the mapping of the recipe to the want texture) and I don't think it needs this much other stuff added. If you want to add botania cleanly to this, imo, the best would be to just replace the two inf plates at each end with elementium caps (I don't think one of each is needed, and weird since most wands don't mix the caps).

EnderProyects:
> I've been thinking on this for a bit. I don't mind the gear changes but I don't like the wand update, personally. I think the recipe is just a bit uglier (misses the mapping of the recipe to the want texture) and I don't think it needs this much other stuff added. If you want to add botania cleanly to this, imo, the best would be to just replace the two inf plates at each end with elementium caps (I don't think one of each is needed, and weird since most wands don't mix the caps).
> 
> Also don't forget to put closes [issue] in the top comment so this links to your issue.

Changed as requested :3

serenibyss (approved):

PR #1085, opened 2024-11-17
Wire rebalance part 2, electric bugaloo or whatever

LazyFleshWasTaken:
Wires suck. There's no rhyme or reason, and loss and amperage and cost are very unbalanced. THis will hopefully bring them to a standard of some kind, and includes a lot of buffs and a few nerfs. Will have a sister pr for the wires in coremod.

Based on https://docs.google.com/spreadsheets/d/1xF-uAQSw-yw2b65Qhip8wnH6RkpEJjr9Ccd4eZM5yUs/edit?gid=911460652#gid=911460652

PR #1194, opened 2025-02-12
Restore Draconium cable/wire loss

ReignOfFROZE:
This undoes part of #1085, because upping Draconium Cables' loss to to 32 EU-Volt per meter/ampere is untenable. This would force everyone to redo their entire wiring system throughout their entire base every tier (not to mention be a MAJOR breaking change for 95% of players when they upgrade to 2.8), as well as make one of the only viable long-term cables exactly not that. 

Even if one tried to argue that at higher voltage tiers EU loss is irrelevant, the GT++ multis tend to use the high end, up to the ENTIRE amp, of voltage for a given tier, and even the shortest of runs from, say, an Active Transformer to a multi would result in frequent powerfails after a short time of usage due to the passive loss on this cable.


ReignOfFROZE:
I don't think that changing this is going to lead to a good user experience, as a lot of players are going to be blindsided by this (unless they read every single line of the patch notes and find this hidden in there) after updating and most of their base nearly immediately falls apart (unless they're putting all of their machines within a few blocks of their power generation/storage), and I'm not alone in this sentiment. 

Refs:
https://discord.com/channels/181078474394566657/181078474394566657/1339029682951032923
https://discord.com/channels/181078474394566657/181078474394566657/1339029487832272926
https://discord.com/channels/181078474394566657/181078474394566657/1339024784977297449

Additionally, LazyFlesh's change wasn't discussed at length anywhere that I can find, and in the PR nobody even commented on the fact that Draconium cables were being nerfed into the ground, so I'm a bit disappointed in the fact that I'm being asked to come up with a counterpoint/replacement concept to be discussed by the wider community for something that was never discussed by anyone outside of the few participants in the original PR (and also affected a LOT more things than I'm attempting to change here), let alone the wider community in an Area of Interest.

I can appreciate that LazyFlesh made other wires viable at various tiers, but I find it to be a bit totalitarian to have the original viable wire be made inviable and then be forced to use a newly viable wire that is now only viable for one tier, and will inevitably have to be replaced every new tier.

Dream-Master:
As I said make a concept with new suggestions we not change a single recipe without a plan .

boubou19:
tbh i'm running through long cable loss issues in my run, that's not going to be pleasant if i do precise cable count just to get my work be ruined by updating 😂. We probably discuss this in depth somewhere in an AOI.

PR #1198, opened 2025-02-13
Buff neutronium and Cosmic Neutronium smelting times

StaffiX:
speeds up both recipes by 20%

|                     **Material with gas**                     | **Before** | **After** |
|:------------------------------------------------:|:-------:|:-------:|
| Neutronium with Xenon | image | image  |
| Cosmic Neutronium with Xenon | image | image |

Also removes non-gas recipe for cosmic neutronium


boubou19:
Neutronium is reduced around ~22.6% and Cosmic Neutronium around 23.2%. Any reason for those values or is it after some trial and error with bart recipes that you concluded it's the closest you can get?

StaffiX:
Could be changed from 384s to 400s for a perfect 20%
For neutronium, it would be 268.8 seconds, i think the 0.8 seconds can be left out for a ~20.2%

serenibyss (approved):

PR #1202, opened 2025-02-17
Buff Crystalline Fluid Conduit Recipe

Ruling-0:
It's just objectively not worth crafting vs adding a bit of pink slime and diamond for the crystalline one. Right now it needs a PTFE pipe, crystalline plate, and 1 ingot of epoxid. One tier down is a tiny PTFE pipe, vibrant plate, and 1 ingot of plastic, and one tier up is a PTFE pipe, pink crystalline plate, and 1 ingot of PTFE. Changes the epoxid to be plastic. It is weaker than the pink crystalline one by half stats, so should be cheaper (thus plastic over PTFE), and I see absolutely 0 reason to nerf any of the higher tier conduits.

Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/19077


Dream-Master (changes requested):
Since we buffed logistic pipes ender io pipes not need a buff they need a nerf all off them

Ruling-0:
> Since we buffed logistic pipes ender io pipes not need a buff they need a nerf all off them

What direction would you want them nerfed in? Cost, tiering, capability? They're already tiered pretty late, and they're pretty expensive to use in bulk.

Dream-Master:
only the recipe. And sure i will recipes for single pipes and bulk pipes. But i not make a concept rn. I just have it on my todo list

PR #1211, opened 2025-02-23
fix wrong recipes in circuit assemblers for preconfigured storage buses

boubou19:
https://discord.com/channels/181078474394566657/939305179524792340/1342457000499613727


boubou19:
tested on our run, works better that way.

PR #1212, opened 2025-02-24
Create new crafting recipes for upgrading transposers with Fluid Regulators

Vlamonster:
Part of [Upgrade Fluid Transfer Rate of OpenComputers Transposers Using Fluid Regulators #19154](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/19154).


Vlamonster:
The recipes are not balanced and should more than likely be changed.

EnderProyects:
Maybe change last tiers to living solder and add Good Generator Fluid storage cores to all the tiers up to 10

Ruling-0:
I think it'd be nice to have a recipe (like shapeless crafting) to break the transposer back into base form and get the regulator back.

Vlamonster:
@Ruling-0 What do you think of adding such a recipe to the extractor? It could return the regulator, but then break the transposer in the process.

Ruling-0:
> @Ruling-0 What do you think of adding such a recipe to the extractor? It could return the regulator, but then break the transposer in the process.

I don't know if that's needed. Many things like this (AE disks, AE CPUs) allow retrieving both parts (this is even similar to AE CPUs which are assembler to put together). I think it'd feel most natural for it to function similarly to those. This is mainly so that if I have a transposer, and need to upgrade it, I can just break it, shapeless craft it apart, then recraft a higher tier one.

src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
```diff
@@ -5617,9 +5619,57 @@ private void makeOpenComputersRecipes() {
...
                         GTModHandler.getModItem(OpenComputers.ID, "cable", 2L, 0),
                         GTUtility.getIntegratedCircuit(1))
-                .itemOutputs(GTModHandler.getModItem(OpenComputers.ID, "transposer", 2L, 0))
-                .fluidInputs(Materials.Plastic.getMolten(72L)).duration(20 * SECONDS).eut(TierEU.RECIPE_MV)
-                .addTo(assemblerRecipes);
+                .itemOutputs(getTransposer(2, 2_560)).fluidInputs(Materials.Plastic.getMolten(72L))
+                .duration(20 * SECONDS).eut(TierEU.RECIPE_MV).addTo(assemblerRecipes);
+
+        GTValues.RA.stdBuilder().itemInputs(getTransposer(1, 2_560), ItemList.FluidRegulator_HV.get(1L))
+                .itemOutputs(getTransposer(1, 10_240)).fluidInputs(Materials.SolderingAlloy.getMolten(72L))
```
serenibyss, line 5626: If you are going to use `getTransposer()` from the script class, then these recipes should also be in the script

Vlamonster: [Move transposer recipes to OpenComputers script](https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/1212/commits/5a7bc9d078dd6438557ddd20aa86e359b30b1638)

serenibyss (approved):

PR #1229, opened 2025-03-31
Adding missing Assembler recipes for transformers

CrystieColon3:
This PR adds missing recipes for uhv+ Transformers, as well as all Hi-Amp and Power Transformers.
Also fixes LV and MV transformers accepting only Copper. Also changes voltage tier of LuV Hi-Amp and Power Transformers to LV, to be in line with all Transformers.
During these changes i noticed inconsistency in UHV tier Transformers, UHV Transformer and Power Transformer use UHV Superconductor Wire, while UHV Hi-Amp Transformer uses Bedrockium. Not sure if this is intended or not, if not i can commit the fix as well. 
Solves https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/19487
java_fZDLkzSVXF
java_nkiuA1BYXM
java_dBTSvGGXo3
java_3bsbZItNjO


KleinGrrmpf:
Thank you. I don't have any experience in modding, but adding recipes could have been something that I could have just C&Ped from other recipes. However, without any coding environment, I didn't want to do that without being able to test it ^^

EnderProyects:
Did you add the lubricant as fluid input in the ones that uses cells?

CrystieColon3:
No, my reasoning was if energy hatches don't keep lubricant as fluid input when swapping to coolant cells, neither should transformers.

EnderProyects:
The current recipes that use lubricant in cells in crafting they also use lubricant when assembled

KleinGrrmpf:
I looked through the commit and it does look like Lubricant is used in Fluid form for the Assembler recipes, yes

EnderProyects:
thats why I deleted the response, but since it says that it doesnt use them...

CrystieColon3:
Oh, you mean do ones that have lubricant cell in crafting table recipe have lubricant fluid in assembler recipe? Then yes, they have

KleinGrrmpf:
Yeah, I was confused by that too ^^

serenibyss (approved):

PR #1243, opened 2025-04-24
New Botania blocks + Tuff Dust

EnderProyects:
Added compresion blocks for pixie dust, mana powder, mana pearl, and gaia ingot
image
image
image
image

image

Added tuff dust
image


EnderProyects:
That was my original idea, but some devs said to add them here instead, if you want to take all from here and make a pr in the botania repo you are free to do it

combusterf:
> what if these were added to gtnh botania instead of NH core mod (for more versatility outside the pack)

Botania is built to be true to the original and portable to other packs. Third party additions typically go here or in Botanic Horizons.

combusterf (approved):
Haven't tested this in-game, but code-wise this looks fine.

PR #1256, opened 2025-05-04
Add uncompressed coal variants to forge hammer recipes

NeOzay:
fix: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/19647

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,7 +1,7 @@
 // Add your dependencies here
 
 dependencies {
-    api("com.github.GTNewHorizons:GT5-Unofficial:5.09.51.263:dev")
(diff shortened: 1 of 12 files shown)


Cleptomania:
Should we also add recipes for the cactus/sugar charcoal/coke blocks(or do those already exist?)

serenibyss:
> Should we also add recipes for the cactus/sugar charcoal/coke blocks(or do those already exist?)

I don't think they already exist, but since those blocks are in GT5u, I think it would be easier to add them there in other sections of existing autogen code for those blocks so I won't hold up this PR

serenibyss (approved):

PR #1260, opened 2025-05-08
Ic2 furnace replace

EnderProyects:
As requested change recipes that use ic2 furnace to use EFR blast furnace

I think I changed all the ones in coremod, but someone shoul check it, since there are many recipes left that I didnt found in coremod
image

### DIFF
--- a/src/main/java/com/dreammaster/scripts/ScriptAutomagy.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptAutomagy.java
@@ -1,8 +1,25 @@
 package com.dreammaster.scripts;
 
+import com.dreammaster.thaumcraft.TCHelper;
+import gregtech.api.enums.ItemList;
+import gregtech.api.enums.Materials;
+import gregtech.api.enums.OrePrefixes;
+import gregtech.api.util.GTOreDictUnificator;
+import net.minecraft.enchantment.Enchantment;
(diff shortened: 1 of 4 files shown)


Ethryan:
Maybe add EtFuturumRequiem.ID to the list of dependencies in each script file too?, its missing from there.

EnderProyects:
thats odd

EnderProyects:
spotless apply messing with me

EnderProyects:
Ah no, is not missing, uh?

Ethryan:
Talking about the public List<String> getDependencies() { in scriptAutomagy on line 40 for example.

EnderProyects:
I have to say that I havent changed those lines in my previous prs and they have worked xd

Ethryan:
Everything works, but usually I like having the mod declared a dependency when used as a part of a recipe.

EnderProyects:
I mean I understan the usage of dependencies in external mods, but in coremod that is only used to change recipes for gtnh it in fact doesnt make much sense, but ok I will add them

Ethryan:
It helps when you are running it in coding environment and not loading the entire pack, without the dependency declaration in each script file everything tends to crash when NEI tries to add recipes for stuff not added to dev environment.

Ethryan (approved):

PR #1263, opened 2025-05-12
Revert the UHT Mesh recipe to the old one

srdr2k3:
The recipe was caught in the crossfire during cleanup, this PR reverts it to the same recipe as in stable.

One day I will stop having ghost commits :)


boubou19 (approved):

serenibyss (approved):

EnderProyects:
I have to say that now that some devs are trying to add super glue to more recipes, superglue could be added to this one ;D

PR #1266, opened 2025-05-18
Don't try to remove null items.

mitchej123:


Caedis:
I was actually debating on adding this check a couple weeks ago

Caedis (approved):

Caedis:
Can you also add it to the other recipe remover methods?

YannickMG:
I see we're not doing any kind of logging of misuses. Is the goal here to encourage calling this method with null arguments as a way to remove isModLoaded complexity on the caller side? I imagine we'd actually want to ensure we almost never call these methods with null arguments.

mitchej123:
I want no exceptions being thrown.   I'm fine with not erroring or someone tracking down the bad removals

PR #1268, opened 2025-05-20
Add recipes for new blocks

EnderProyects:
Add gt machine recipes for mana powder block, mana pearl block, pixie dust block, and gaia ingot block, and its decompression recipes
Needs https://github.com/GTNewHorizons/GT5-Unofficial/pull/4281 since it uses prismatic acid to turn gaia ingot block back into gaia spirits

image
image
image
image
image
image
image
image


EnderProyects:
@combusterf btw using this we can add a bulk recipe for terrasteel using the full blocks variants instead of single items, what its your opinion about this?

PR #1276, opened 2025-05-24
Reduce Crafting time of lootbag conversion

Nockyx:
These recipes were added to upgrade lootbags is a way better than repeatedly clicking on the quest submission on questbook, but the recipes are all way too slow and it is considerably faster to just keep trading them via questbook instead.

This PR reduces the recipe duration so these recipes aren't useless and people don't continue to just manually press the quest submission button repeatedly.


Dream-Master:
Make it 10 secs like the coins. 30 sec us indeed long but 1 sec us a joke

Nockyx:
> Make it 10 secs like the coins. 30 sec is indeed long but 1 sec us a joke

when i tested in game it takes like 3 seconds to get it via the quests, but fine

Dream-Master (approved):

PR #1278, opened 2025-05-27
Change T1 air filter controller recipe to use proper casing

serenibyss:
Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/19696

Changes the T1 air filter controller to use an air filter turbine casing rather than a normal turbine casing, matching the structure of the multiblock


Ethryan (approved):

Pilad:
Maybe the code below should be changed to something similar?

Casing_Turbine2 to
    Casing_AirFilter_Turbine_T2 for Machine_Multi_AirFilterT2
Casing_Turbine3 to
    Casing_AirFilter_Turbine_T3 for Machine_Multi_AirFilterT3

PR #1291, opened 2025-06-15
Move Custom1-5 Aspect creation to GT5U

OmdaCZ:
Part 2 of https://github.com/GTNewHorizons/GT5-Unofficial/pull/4473 
Will undraft after testing, whether existing instances of custom aspects cause issues.


koolkrafter5:
You should also delete the textures here since they were moved to gt5u

PR #1292, opened 2025-06-18
Zero point module Recycling

EnderProyects:
image
It says that you can recycle them, but you cant, add a recipe to get some funny weird materials
image

My main idea is to add chionite into the Stargate-Crystal Dust instead of the Omega Baryon

### DIFF
--- a/src/main/java/com/dreammaster/scripts/ScriptSGCraft.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptSGCraft.java
@@ -3,26 +3,35 @@ package com.dreammaster.scripts;
 import static gregtech.api.enums.Mods.AE2FluidCraft;
 import static gregtech.api.enums.Mods.AppliedEnergistics2;
 import static gregtech.api.enums.Mods.BartWorks;
+import static gregtech.api.enums.Mods.DraconicEvolution;
 import static gregtech.api.enums.Mods.EnderIO;
 import static gregtech.api.enums.Mods.GalacticraftAmunRa;
 import static gregtech.api.enums.Mods.GoodGenerator;
 import static gregtech.api.enums.Mods.OpenComputers;
 import static gregtech.api.enums.Mods.SGCraft;
 import static gregtech.api.enums.Mods.TecTech;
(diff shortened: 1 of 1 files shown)


Ruling-0 (changes requested):
Without speaking to this feature at large, certainly it should not give SGCraft Naquada, that is purposefully useless and recipeless.

Dream-Master:
how rare is this item?

EnderProyects:
I haven't found a single ZPModule as loot, ( you can get one at a secret zpm quest)

EnderProyects:
> This need more discussion.

About the shards part, or the chionite?

LazyFleshWasTaken:
While the "funny" item selection seems a bit, uh, diverse for something supposedly made by a long gone advanced civilisation (why amber?) it does look like an alright recycling recipe. 

The modules are both rare enough, and the rare materials output low enough that it shouldn't pose a big balance issue, which I think would be the main concern of adding it. I forget if you can find it before UV/UHV, in which case you can't even use the awdrac or chaos shards, though the dilithium you can make a lens out of for something in the nano forge iirc. 

Can someone confirm the lowest tier it can be found (and how much trouble it would take TO find it)? 
If the standard way to find it isn't below UV/UHV, I think it's a fine addition.

Dream-Master:
Yes, I think the things you get from it shouldn't be so ordinary, but rather something special. To me, it looks like it was thrown together randomly without a concept (or maybe I didn't understand the concept).

EnderProyects:
I didnt want to place rarely stuff so it doesnt affect balance, the materials I placed where more by the colours of the original Zero Point module
image

EnderProyects:
Then any idea? I want to maintain the chionite and the shards, the other 7 materials left can be whatever

LazyFleshWasTaken:
Actually, another option is that it directly regenerates an island at 10k 10k, which may actually be the simplest option, if a bit destructive.

EnderProyects:
Change the output from chaos shard to chaotic dust, it will be used to craft a device to track chaos islands @GDCloudstrike 
image

PR #1296, opened 2025-07-02
Remove access to GTMod isclient methods

Alexdoru:
This PR removes a dependency on GT methods that are being deleted,
It must be merged and build before the linked PR. https://github.com/GTNewHorizons/GT5-Unofficial/pull/4541

### DIFF
--- a/src/main/java/com/dreammaster/main/MainRegistry.java
+++ b/src/main/java/com/dreammaster/main/MainRegistry.java
@@ -155,7 +155,7 @@ public class MainRegistry {
      * Returns true on a client
      */
     public static boolean isClient() {
-        return proxy instanceof ClientProxy;
+        return proxy.isClient();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/com/dreammaster/main/MainRegistry.java
```diff
@@ -151,6 +151,20 @@ public static void AddLoginError(String pMessage) {
         }
     }
 
+    /**
+     * Returns true on a client
+     */
+    public static boolean isClient() {
+        return proxy instanceof ClientProxy;
```
serenibyss, line 158: Why not put this in the proxy class like GT did? I think this could cause problems since ClientProxy implements an interface marked as client only

Alexdoru: that's possible, I will change it

serenibyss (approved):

PR #1304, opened 2025-07-17
Remove tunstensteel from MM MK0

RecursivePineapple:
The MK0 should be craftable in HV with some branching out to other mods (EIO, TC, etc). It shouldn't require an EV/IV material like tungstensteel. This reduces the field generator from MV to LV to fix this oversight.


Dream-Master:
What is MM and MK0?

chrombread:
> What is MM and MK0?

Matter Manipulator, Mark 0, the primitive one(I think)

chrombread (approved):
Looks good

PR #1305, opened 2025-07-18
Changes EU/t of Bio SOC recipe to be in line with other circuits

chrombread:
discussed in meta-dev, this will allow Bio SOC in quad CAL setups to share the energy hatch, instead of making it share one ehatch per 2 cal's (or 1:1 if lazy). this makes mainframes the only non standardized circuit, which can be left unchanged or for a separate PR.

image


boubou19:
please post the link of the message from #meta-dev, so we can give a reference to the discussion, for project history

PR #1307, opened 2025-07-22
Adds Tumbaga Recipes to Alloy Smelter

UltraProdigy:
## Changes:
- Does exactly as title says, adds 4 new recipes, so the ingots and dusts can be used in any combo for creation as is consistent with other alloy smelting recipes
- Added at a slight time increase but same EU & ratio to the mixer / furnace combo since that appears to be status quo with other alloy smelter recipes.

There is no reason that Tumbaga should be omitted from these recipes, as far as I can tell it was omitted because it is a GT++ alloy so it didn't fall in line with the group code in GT5u.

### In-Game Photo (1 of 4 Recipes):
image


LewisSaber (approved):
lgtm

PR #1310, opened 2025-07-24
Convert Travellers Gear armor stands to a chest

koolkrafter5:
Chest contains the stand's items and a Bibliocraft armor stand (they were interchangeable before TG was removed).
Only happens when TG is not loaded.

Example (TG items aren't transferred, obviously):
image
image


koolkrafter5:
Marking as draft because I realized I should convert the items, too

koolkrafter5:
Converting the items needs some placeholder item, which I don't know how to create

PR #1315, opened 2025-07-26
Gets Rid of Duplicate Tumbaga Recipes

UltraProdigy:
### Changes:
As title states, there were duplicate recipes for tumbaga gear, frame box, and rod. I assume they were put their initially as supplemental but they are no longer needed. 

This will fix and close https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/20602


Nikolay-Sitnikov (approved):
Looks like a simple fix, although I can't confirm that the bug exists in the first place.

PR #1316, opened 2025-07-27
EFR Copper Item Recipes

UltraProdigy:
## Changes:
- Adds copper trapdoor recipe for craft & assembler (based on iron trapdoor)
- Adds copper door recipe for craft & assembler (based on iron door)
- Adds oxidation recipes for all copper blocks, doors, trapdoors, & slabs, including a way to reverse the process
- Adds recipes to un-wax the waxed copper items
- Adds missing assembler recipes for cherry trapdoors

This is continuing the chain of EFR recipe addition PRs, all tested in latest of course.

### Copper Trapdoor:
image
image

### Copper Door:
image
image

### Oxidation Recipe Example (Extrapolate to all in list):
image
image

Works up tier by tier, regular makes exposed, makes weathered, makes oxidized in the chemical reactor/LCR, and works back down the ladder in the same way in the arc furnace.

Recipes are based on real chemistry (with a little grace of course): 2Cu + O2 + 2CO2 -) 2CuCO3 (oxidized copper). Then stripped of oxygen by way of large amount of heat and Hydrogen CuCO3 + H2 -) 2Cu + CO2 + H2O, where the CO2 and H2O are evaporated back into the air via the arc furnace.

### Un-Waxing Recipe Example (Extrapolate to all Copper Items):
image

Works in CR/LCR, fluid is acetone, something you'd expect to rid wax with.

### DIFF
--- a/src/main/java/com/dreammaster/scripts/ScriptEFR.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptEFR.java
@@ -100,6 +100,8 @@ public class ScriptEFR implements IScriptLoader {
                         GTModHandler.getModItem(ExtraUtilities.ID, "budoff", 1, 0), 'D',
                         GTModHandler.getModItem(Minecraft.ID, "comparator", 1, 0), 'E', "gearGtSmallAnyIron" });
 
+        // Cherry Trapdoors
+
         GTModHandler.addCraftingRecipe(
                 GTModHandler.getModItem(EtFuturumRequiem.ID, "cherry_trapdoor", 1L),
                 bits,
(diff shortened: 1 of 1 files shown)


UltraProdigy:
> @UltraProdigy i'm not sure hydrogen can oxidize anything from the small chemistry i remember.

Correct, hah. I'm "skipping a step" here. The arc furnace recipe is taking away the green look of the copper (reducing). So with intense heat the CuCO3 will decompose into CuO and CO2, and then still under heat when presented with gaseous H2 the CuO will reduce copper back to pure metal and release excess water into the air.

boubou19:
Well if it makes sense, then recipes code looks fine by me

koolkrafter5:
The amount of gas used seems a bit high for making decorative building blocks.

boubou19 (approved):

PR #1319, opened 2025-07-29
Add Tipped Arrow & Cherry Plank Recipes for EFR

UltraProdigy:
## Changes:
- Adds Cherry plank recipes (crafting & cutting, consistent with other plank recipes) - oreDict will also be added to everything as my last recipe PR
- Adds Tipped Arrow recipes, with a more beneficial assembler option to provoke users to use that recipe instead of traditional crafting.
- Also does one more follow up fix to https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/1318, apparently this effected the other types of tank carts as well, luckily all fixed by one number.

The tipped arrows are made with lingering potions, those recipes will be coming in a GT5 PR later. 
This will officially close https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/20264

### In-Game Photos
image
image

** This applies to all 32 arrow types

image
image

** The saw and log are switched now, screenshot was taken before, also there are 3 cutting machine recipes two aren't depicted


chrombread (approved):
lets get this shmoney

2ndDerivative:
what exactly was this fixing with changing the railcraft tanks?

PR #1320, opened 2025-08-01
EFR Shulker Recipes

UltraProdigy:
## Changes:
- Adds arcane crafting recipe for all 8 variants of shulker
- Adds arcane crafting recipe for all 9 variants of shulker box upgrade
- Adds those arcane crafting recipes to the Shulker page in the Thaumonomicon 
- Adds shapeless dyeing & undying recipes for every type and color of shulker box (Allows inventory of shulker box to be preserved between each craft)
- Removes recipe for cherry slab, code was moved to GT5u earlier so this was a duplicate.

### Shulker Boxes (In Order of Size)
Plates are living wood, living rock, and reinforced glass for crystal. Shulker types correspond to barrel type

image
image
image
image
image
image
image
image

### Shulker Box Upgrades:
All are straightforward like first photo, single plate in center with previous material surrounded by double plates (except diamond, that doesn't exist) with upgraded material. The Vis cost associated is the exact difference in cost between the barrel made and being upgraded to. The second photo is the only exception, for crystal and obsidian being a flat 5 vis cost despite technically having no difference between upgrades.

image
image

### Shulker Dyeing & Un-dyeing:
Any shulker of any color or type can be dyed and it will maintain it's type and maintain all items in it's inventory. For now they can be undyed by simply shapelessly crafting them by themselves.

image
image

### DIFF
--- a/src/main/java/com/dreammaster/scripts/ScriptEFR.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptEFR.java
@@ -1264,7 +1264,7 @@ public class ScriptEFR implements IScriptLoader {
                 'd',
                 getModItem(EtFuturumRequiem.ID, "diamond_barrel", 1, 0, missing),
                 'e',
-                GTOreDictUnificator.get(OrePrefixes.dust, Materials.Titanium, 1));
+                GTOreDictUnificator.get("dustCrystallinePinkSlime", 1));
         ThaumcraftApi.addArcaneCraftingRecipe(
                 "SHULKER",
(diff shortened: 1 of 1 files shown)


EnderProyects:
Yea, you can just set all to crystalline alloy, so there is no difference at using compressed chests

UltraProdigy:
Okay we should be all set. Made all shulkers take the crystalline alloy, changed the gold -) diamond upgrade back to salis mundus since there is no bridge to gap anymore.

Dream-Master (approved):

PR #1325, opened 2025-08-09
Add deepslate dust

RecursivePineapple:
image
<br />
image
<br />
image
<br />
image


RecursivePineapple:
The lack of chances are intentional, I'm not sure why tuff has them because standard stone doesn't and these outputs aren't anything special (like the rare drops from space stone dust).

PR #1329, opened 2025-08-09
refactor infusion recipes to use oredict strict variants from tc4tweaks api

Glease:
Fix https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14631 and maybe some other instances.

In case there are other cases where implicit oredict can be problematic, I did an semi-automatic refactor to refactor all infusion recipes defined by this mod to
1. If it's clear that the ingredient is something from oredict, i.e. GTOreDictUnificator.get(), it's now specified to use that oredict name explicitly.
2. Otherwise, the new recipe will accept the original item passed in **only**, and its oredict substitutes (if any) will be rejected.

I did not look at every recipe modified and there might be cases where oredict would have been desired and I'm ready to correct them as soon as possible.

This changes all infusion recipes defined by coremod to use the new InfusionRecipe subclass from [TC4RecipeLib](https://github.com/Glease/TC4RecipeLib).
Quoting its README,

> This library provides an InfusionRecipe subclass with enhanced oredict handling capabilities.
>
> The basic InfusionRecipe class offered by thaumcraft itself uses a broken oredict handling algorithm.
It is broken in these ways:
> 1. Unconditionally and implicitly do oredict on all inputs. For instance, If your recipe require a specific type of wood (such as warpwood from Tainted Magic), thaumcraft will automatically allow any other kind of wood as well.
> 2. Only considers the first oredict name of the specified stack, and only check if any oredict name of player supplied ingredient is that first oredict name. This is problematic when items have more than one oredict name.
>
> This library, on the other hand, require mod author to give explicit instruction on whether oredict is needed.
It unifies various kinds of input under one interface RecipeIngredient.
Library itself provides a few implementation to handle common use cases and you should not need to implement your own subclass.

For more information, please refer to its complete README.

Related to https://github.com/GTNewHorizons/TCNEIAdditions/pull/32 and these 2 should be merged together.

@Dream-Master if this is to be merged, please be sure to upgrade TC4Tweaks to 1.5.37

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -7,6 +7,8 @@ dependencies {
 
     implementation("com.github.GTNewHorizons:GTNHLib:0.6.38:dev")
 
+    implementation("net.glease:tc4recipelib:1.5.37:dev")
+
     compileOnly("com.github.GTNewHorizons:AkashicTome:1.2.6:dev") { transitive = false }
     compileOnly("com.github.GTNewHorizons:Avaritia:1.72:dev") { transitive = false }
     compileOnly("com.github.GTNewHorizons:Irontanks:1.4.2:dev") { transitive = false }
--- a/repositories.gradle
+++ b/repositories.gradle
@@ -14,4 +14,15 @@ repositories {
             artifact()
         }
     }
+    exclusiveContent {
+        forRepository {
+            maven {
+                name = 'glease'
+                url = 'https://maven.glease.net/repos/releases/'
+            }
+        }
+        filter {
+            includeGroup('net.glease')
+        }
+    }
 }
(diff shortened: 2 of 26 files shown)


UltraProdigy (dismissed):
Actually I apologize, I wasn't thorough enough on my first round. I found an issue by playing around more. There is an "error during loading recipe" anytime 'U' is clicked in NEI, I imagine because one of the registered infusion recipes is null somewhere, it's not game-breaking but should be addressed. I can start sifting through to find it.

EnderProyects:
Usually nulls are replaced with fire, so search for recipes for or using fire, and maybe you find it

UltraProdigy:
Yeah I was hoping it would be that easy, but it's hidden. There are the same 2003 recipes, and when I combed through before all had valid items & no recipes appeared broken or nulled & infusions working fine which is why I initially approved. Trying to have log output error now.

UltraProdigy:
Okay here's the shtick. It wasn't exactly a null recipe but the fact that addInfusionCraftingRecipes can pass along null components arrays to tc4tweaks which causes the error when it's streamed from there via the nei "U" function. Additionally there was a second issue getting npe's from calling raw.getAspects(). 

I currently have a solution that makes everything function nicely, essentially in coremod I made it so the components array & aspects are never null and theres additionally an edit to TCNEIAdditions that avoids calling raw.getAspects & skips bad entries.

But I don't know if it's the most elegant solution, (though it is safe) it's possible that this would be better handled by an adjustment to tc4tweaks in the first place. So whether you want to implement a solution yourself or take mine I'm game either way.

UltraProdigy (approved):

Glease:
@UltraProdigy I've fixed it at recipelib level but IMO your fix here can be retained as a safety measure

PR #1331, opened 2025-08-12
made deepslate viable

seventh-june:


boubou19 (approved):

boubou19 (dismissed):
It appears that the chemical rules aren't respected (more info at https://wiki.gtnewhorizons.com/wiki/Chemical_Balance)

Pxx500:
1) recipes are not properly chembalanced (inputs must equal outputs, both items and fluid buckets count)
2) deepslate is basically free to make (rock breaker with magma block no consume) so making this create rare resources such as boron is gonna trivializea lot of things
 a) mica electrolyzes for fluorine which would make it too trivial
 b) maybe boron trioxide instead of raw boron?
4) aluminium oxide instead of aluminium in staurolite/epidote

Perhaps gate it higher and require chemicals (acids, bases or others) to perform the deepslate refining

seventh-june:
@Pxx500 fixed this in latest pull, chemical formulas are balanced, boron is now a smaller output of datolite (as it was supposed to be), and electrolyzer recipe for it was nerfed to EV. Also replaced mica with plagioclase as one of the larger byproducts of deepslate dust, which is still found in metamorphic rocks.

LazyFleshWasTaken:
Needs the gt5u one to be merged first.

Pxx500 (approved):
everything fixed

Pxx500:
(might need to nerf later, we'll see)

PR #1335, opened 2025-08-15
Shulker and qol botania

EnderProyects:
Add a mixer recipe to get gaia dust instead of using the molecular ingot recipe
image
Add a recipe to get shulker eggs and shells
image
image

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/MixerRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/MixerRecipes.java
@@ -3,6 +3,7 @@ package com.dreammaster.gthandler.recipes;
 import static gregtech.api.enums.Mods.AppliedEnergistics2;
 import static gregtech.api.enums.Mods.Avaritia;
 import static gregtech.api.enums.Mods.BiomesOPlenty;
+import static gregtech.api.enums.Mods.Botania;
 import static gregtech.api.enums.Mods.Chisel;
 import static gregtech.api.enums.Mods.DraconicEvolution;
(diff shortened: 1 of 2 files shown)


combusterf:
- We're supposedly in a feature freeze
- There is no balance discussion
- Only images are posted, and the recipes can't therefore be reliably determined
- Spawn eggs in particular are very hard to obtain outside the questbook in the current version, so labeling this as QOL is misleading and this recipe in particular is probably way too easy.

EnderProyects:
All the discussion was made on the discord

EnderProyects:
The coocon is ev, so is well gated after shulkers, the idea is to use that recipe in order to make EEC if shulkers arent added to world generation, since there is already a quest to obtain the egg

combusterf:
But did you actually discuss it, or even consider the other consequences of making spawn eggs craftable?

EnderProyects:
Just check the discord, I think it will be easy to spot since it will tell you message by blocked user, or something like that... Also what are the consequences of making spawn eggs crafteable? You can get a ton at the same tier using EEC, a part of that, the Gaia recipe is a big qol, and the shuker shell recipe is just if someone doesn't whant to kill one shulker each hour the quest for the egg updates

combusterf:
To answer your other questions:

1: Spawn eggs are used to craft dragon eggs
2: Spawn eggs in the last version are obtainable through:
  - Questbook
  - EECing some special boss mobs (you'll have to find and catch them, vanilla mobs do not work)
  - Silverfish bees.

EnderProyects:
Check the recipes for the Big Egg, I think you missed a change that was made some months ago

PR #1337, opened 2025-08-17
bring tantalum freeze time to a reasonable level

Pxx500:
Tantalum was a major outlier in freezing time (longer to cool than HSSG or TungstenSteel ingots), this PR brings more in line. MV chembatch recipe untouched


boubou19 (approved):

boubou19:
For completeness, here is the chembath recipe: 
image

PR #1338, opened 2025-08-18
Adds Colored Bed Recipes

UltraProdigy:
### Changes:
- Adds recipes for EFR colored beds; the recipes (both crafting and assembler) are exactly the same but differentiated on the color of carpet used in the recipe. All types of wooden planks are still compatible 
- Modifies the regular minecraft bed recipe to account for the carpet coloring

There are a lot of deletions because there were an obscene amount of bed recipes that could've been easily taken care of with a few lines and a for loop. Tested in game ofc

Frankly, I don't like the overabundance of NEI recipes for all sorts of items nor the code spam I see so often due to different types of inputs so I will very much be targeting things like this in a refactor for 2.9.

This will have a sister PR modifying the bed quest to take the OreDict version so any bed craft will complete it and not just the red one.

### Recipe:
** Except the plank / fence type can be any, and changing the carpet gets you a different colored output

image
image


Caedis:
Update your EFR config
etfuturum/functions.cfg

    # Render beds with a 3D inventory model instead of a 2D sprite. [default: true]
    B:inventoryBedModels=true

PR #1341, opened 2025-08-21
Remove Silver melting since there isn't any silver casting

LazyFleshWasTaken:
The issue comes back out every once in a while, but nothing happened, so here you go. Lead doesn't have casting, and the same issue was fixed by removing its melting.

fixes [#20480](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/20840)
fixes [#16044](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/16044)


UltraProdigy:
Tested in game I assume?

LazyFleshWasTaken:
> Tested in game I assume?

...no..., but its a pretty simple change so It shouldn't break? I'll test it rn ig

LazyFleshWasTaken:
Just tested, it has 1 recipe, gear to fluid, and then fluid to gear casting, but lead has the same, so I assume they're removed somewhere else.

PR #1342, opened 2025-08-22
Add recipe for crushed obsidian

Ableytner:
grafik

As crushed obsidian only has a blast resistance of 27 and isn't used in any recipes, I tiered it to LV.

Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/19369


lynxx131 (approved):
Looks fine to me

PR #1343, opened 2025-08-24
Adds Recipe for Super Stock Replenisher & Makes Void Cell Recipe Consistent

UltraProdigy:
### Changes:
- Adds in recipe for super stock replenisher, UV gated since it performs UV gated actions (stocks fluids)
- Changes void cell recipe to more closely mimic the fluid void cell recipe (replaced ender eye with ae2 singularity)

(There will also be a quest for SSR it's coming in a sweep probably tomorrow)

### Super Stock Replenisher Recipe:
image
image

### Void Cell Change: (Void Cell on Left, Fluid on Right)
image
image


chrombread (approved):
lets get this shmoney

PR #1345, opened 2025-08-24
Adds controller recipe for steam alloy smelter

EnderProyects:


sisyphussy:
does this also have an assembler recipe? (not necessary but would be nice)

EnderProyects:
No, none of the steam multis has assembler recipes

UltraProdigy:
I think the frame box & pipes should maybe be Tumbaga to be consistent with the other steam multis?

sisyphussy:
Alright, understood. However, your code is in the wrong place. The recipe needs to be initialized during the postInit phase. https://github.com/GTNewHorizons/GT5-Unofficial/blob/6f40a0564b57a104f3fd50a11ebd8002d236e4f2/src/main/java/gtPlusPlus/core/util/minecraft/RecipeUtils.java#L100

sisyphussy:
> I think the frame box & maybe pipes should be Tumbaga to be consistent with the other steam multis?

I agree with the casings, but I don't think tumbaga has pipes. Bronze pipes are fine.

UltraProdigy (approved):

sisyphussy:
is there a quest for this multiblock yet?

EnderProyects:
> is there a quest for this multiblock yet?

idk, if you whant to add one go for it I guess

PR #1350, opened 2025-08-31
netherite dupe fix

chrombread:
fixes https://github.com/GTNewHorizons/Dupes-Exploits-GTNH/issues/190
Now:
image

ALSO:

includes a devOnlyNonPublishable that allows the dev-env to be booted in the first place. a recent pr to gt5u included a devonly dependency. If there is a better route that can be taken with this, please let me know and ill revert it.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,7 +1,7 @@
 // Add your dependencies here
 
 dependencies {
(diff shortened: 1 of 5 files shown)


LewisSaber (changes requested):
time should be x2

PR #1352, opened 2025-08-31
Equalise Botania-TC integration with Base Thaumcraft

LazyFleshWasTaken:
Botania is a magic mod that isn't very integrated yet, and feels like a side project with rewards solely within itself, unlinked back to the other magic mods.

In sight of that, I want to decrease the nerfing of botania tc-wand stuff, like the livingwood core and manasteel cap, two early-game core/caps that are equivalent to thaumium and silverwood respectively, and should therefore hold a similar, if not cheaper cost to reward branching into botania.

Make no mistake, while this is making the recipes cheaper, it is only doing so to bring parity to their 'equals' in the main TC mod. 

Starting with the cores, here are the current recipes for the wands 'equal' to silverwood, and livingwood's upgrade, dreamwood:
image
You may notice that the livingwood core requires all the elemental runes, mana, and season runes, which is far more expensive (and a much lenghtier fetch quest, but that's a gripe with base botania, not the integration) compared to the crystal clusters of silverwood. It then also needs to be thrown in a mana pool. The aspects, at least, are fine. The wand below silverwood is witchwood, which just needs mutandis, a wtiches oven, and a bunch of alchemy to make vinteum, all doable with just a bit of witchery.

All have 100 vis capacity.

Dreamwood is the upgrade to livingwood, yet it has literally the same recipe except for a different block in the centre.

My proposed recipe is : 
image

This removes the season runes, bringing it more in line with the solely elemental-themed inputs of silverwood, and leaving dreamwood to take both, as an upgraded wand. Frankly, i'd take off the elemental runes for dreamwood and leave it with just the season ones, but I didn't think of it until just now.

Manasteel caps are 'equal' to Thaumium. I mean, Vinteum too, but they're a shortcut to Thaumium and so I feel their recipe is fine. Anyway, base recipes: 

image

Not only does manasteel take an extra infusion, its infusion cost is increased. This is for an 'equal' to thaumium, which has 1 arcane crafting and 1 infusion. Manasteel begins as inert silver, then the silver caps have to be charged with an infusion, then infused again to transform to manasteel, to then be thrown in a mana pool to be charged.

Again, this is a cap equal to thaumium.

My proposed recipe is such: 
image

First, remove 1 infusion by transforming an _inert_ silver to an inert manasteel, then bring the infusion cost to be in line with thaumium, since astral silver is a mix of silver and thaumium, and lastly, change silver caps to require steeleaf screws, to be in parity to how thaumium require knightmetal screws. Frankly, steeleaf is more annoying to get than knightmetal, but that may just be my opinion. Also, if you can get steeleaf, you can get aluminium, hence I don't feel that it makes that much difference between the two.

And once again, while this is making the recipes cheaper, it is only doing so to bring parity to their 'equals' in the main TC mod.

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/AssemblingLineRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/AssemblingLineRecipes.java
@@ -61,7 +61,6 @@ import gregtech.api.util.recipe.Scanning;
 import gtPlusPlus.core.block.ModBlocks;
 import gtPlusPlus.core.material.MaterialsAlloy;
 import gtPlusPlus.core.material.MaterialsElements;
-import gtPlusPlus.core.material.Particle;
 import gtPlusPlus.xmod.gregtech.api.enums.GregtechItemList;
 import gtnhlanth.common.register.WerkstoffMaterialPool;
 import tectech.recipe.TTRecipeAdder;
@@ -1361,10 +1360,13 @@ public class AssemblingLineRecipes implements Runnable {
                     .eut(TierEU.RECIPE_UIV).addTo(AssemblyLine);
 
             // Quark exclusion casing
-            GTValues.RA.stdBuilder().metadata(RESEARCH_ITEM, Particle.getBaseParticle(Particle.STRANGE))
-                    .metadata(SCANNING, new Scanning(40 * SECONDS, TierEU.RECIPE_UHV))
-                    .itemInputs(
-                            GTOreDictUnificator.get(OrePrefixes.frameGt, Materials.SuperconductorUHVBase, 32),
+            TTRecipeAdder.addResearchableAssemblylineRecipe(
+                    ItemList.StableBaryonContainmentUnit.get(1),
+                    512 * 40 * SECONDS,
+                    512,
+                    (int) TierEU.UHV,
+                    4,
+                    new Object[] { GTOreDictUnificator.get(OrePrefixes.frameGt, Materials.SuperconductorUHVBase, 32),
                             GTOreDictUnificator.get(OrePrefixes.frameGt, Materials.Ledox, 32),
                             GTOreDictUnificator.get(OrePrefixes.frameGt, Materials.CallistoIce, 32),
                             GTOreDictUnificator.get(OrePrefixes.frameGt, Materials.EnrichedHolmium, 32),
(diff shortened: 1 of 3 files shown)


Dream-Master:
I am not sure buffing all stuff is a good solution . You remove also complex recipes and simplifying it. Balancing is good but not at any price . I will check the or later ingame

combusterf:
My personal take on this subject is that I'd personally rather see the WG wand items have capacities different from the default progression. That rewards users by allowing less intermediate wands to be made before obtaining core progression goals (like the silverwood scepter) or getting there with less wand hacking enchants. 

Also, the base ingredients come with tiering restrictions. dreamwood/terrasteel/elementium come relatively late in Botania progression so "buffing" the recipes overlooks that you can't make them anyway. Again, a better argument to buff their base stats instead.

Dream-Master:
> My personal take on this subject is that I'd personally rather see the WG wand items have capacities different from the default progression. That rewards users by allowing less intermediate wands to be made before obtaining core progression goals (like the silverwood scepter) or getting there with less wand hacking enchants.
> 
> Also, the base ingredients come with tiering restrictions. dreamwood/terrasteel/elementium come relatively late in Botania progression so "buffing" the recipes overlooks that you can't make them anyway. Again, a better argument to buff their base stats instead.

so you suggest buff wand values instead of recipes?

LazyFleshWasTaken:
> My personal take on this subject is that I'd personally rather see the WG wand items have capacities different from the default progression. That rewards users by allowing less intermediate wands to be made before obtaining core progression goals (like the silverwood scepter) or getting there with less wand hacking enchants.
> 
> Also, the base ingredients come with tiering restrictions. dreamwood/terrasteel/elementium come relatively late in Botania progression so "buffing" the recipes overlooks that you can't make them anyway. Again, a better argument to buff their base stats instead.

The problem with this, is that if they're buffed enough to go into another tier, they'll be gated again. If they're barely buffed, but retain the same recipe, there'd be no point, because no one is going to make them anyway.

That's why I want to make the cost relatively equal to the parallel item in base thaum, the reward of special abilities coming because we want to encourage branching out to the other magic mods. Witchwood has it down fairy alright, if a bit of an annoyance with the vinetum since you have to make 81 nuggets of it to make a block to then hammer it down into gems, as it is both an ingot and a gem. Or mine it from Mars, ig.

This pr only touches livingwood rod and manasteel caps, because I wasn't sure if I wouldn't be immediately dismissed if I did all of the cores at once. The blood magic rods and the tainted rod are also far more annoying-- sure it was an attempt at making it expensive/complex, but it does little more than have the person check it out, blanch at the recipe, and dismiss it entirely-- to craft, one because it needs _all_ the alchemical essences, one of which needs titanium iirc, and the other because it needs you to mine tainted stuff with the shove of the purifier, which for some reason needs a silverwood rod and exquisite gems now.

LazyFleshWasTaken:
ah fuck, i just realised astral silver is an hv mixer recipe. Gonna need to change that before merging, or its gonna remain hv for manasteel caps.

hm... no silver material looks appropriate, so would one thaumium dust, 2 silver dust be fine? or 2 silver 1 salis? No, wait, manasteel is already a magic metal, and silver caps are theoretically a vis-charged metal already, so then just 3 silver dust? That's not really any more expensive than 3 thaumium dust. 3 salis to match thaumium caps is the other option ig.

Nikolay-Sitnikov:
> My personal take on this subject is that I'd personally rather see the WG wand items have capacities different from the default progression. That rewards users by allowing less intermediate wands to be made before obtaining core progression goals (like the silverwood scepter) or getting there with less wand hacking enchants.
> 
> Also, the base ingredients come with tiering restrictions. dreamwood/terrasteel/elementium come relatively late in Botania progression so "buffing" the recipes overlooks that you can't make them anyway. Again, a better argument to buff their base stats instead.

My take on this is that it would be better to un-gate dreamwood & elementium, since everything they gate is completely worthless at IV anyways. (Terrasteel Caps are clearly intended as a joke item - they don't even get an infusion recipe - so they're supposed to be useless.)

LazyFleshWasTaken:
Now that feature freeze is over, can we continue discussion on this?

boubou19:
So what do we do with this PR?

LazyFleshWasTaken:
> So what do we do with this PR?

Dreammaster told me in my dm's that combuster said it needs some steering to get to a consensus before it can be finalised. 

I've been relatively busy, so I haven't pushed for much discussion.

PR #1354, opened 2025-09-02
Integrate EFR in draconic evolution, I think it suits well in this recipes

EnderProyects:
Will post images when I finish downloading latest daily ( bad internet)


EnderProyects:
It makes the armor a bit more cheap  since it removes some draconium, and the materials it use are some you may already have since the netherite line and shulker EEC, the chorus can be found in the end and there is also a TGS recipe for it, so it just incentive some exploration ( and there are also recipes for Gog or special runs so all parts can still be crafted)

Nockyx:
low tier materials on uhv recipes, i dont see any reason for this pr

Ghostipedia:
> I agree with Noc here, I think this needs to either have massive changes made or to be closed entirely. The recipes don't really make much sense, and they act as a large buff. To me, it just feels like we tried to throw a bunch of random EFR & Netherite line stuff into the recipe just for the point of cross mod integration and it comes across as sloppy.

Want to toss my two cents into the ring as a packdev who handles cross-mod integration with GT on the daily with my own personal project.
These are what I'd like to refer to as "Divine Journey" recipes, where the logical method to get to them is based on what the properties of the items are, verses their actual technological progress. Which imo is not what NH does.

That said, I don't think EFR integration into DE is sensible or reasonable *at all*.
Why? Drac Evo in NH is something very late into the pack and the only think i could even see justified here is the Netherite Material post ZPM. But even that is a stretch.

I say this kindly, but as an outsider there's two methods of approach here
A. Recipes that make 'logical' sense based on item properties (the approach above)
B. Recipes that make 'progression' sense based on materials near the unlock point.

I strongly second the suggestion UltraProdigy mentions above with using 'higher tier' parts. But no mundane basic material directly from EFR should be used. 

Another thing I'm seeing is parts in these recipes that feel a bit shoehorned, can you give more reason why there's a lamp being jammed in a helmet when it'd more logically make sense to be NVG?? 

Apologies if this comes off strong, this is just a change I can't seem to get behind atm.

PR #1358, opened 2025-09-03
Adds Missing Lootbag Assembler Recipes (& Light Refactor)

UltraProdigy:
### Changes:
- Refactors a bunch of repeating code so that it's relatively easy to enter in new lootbag recipes if we introduce more, it should be very easy to add new pairs of bags and change quantity or voltage for balance reasons (I might do the same for the coins)
- Adds new recipes for AE2 bags, witchery bags, and HEE bags that weren't there before

This can be merged at any time safely, but is meant as a sister PR to https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/21067


LewisSaber (approved):
LGTM

PR #1363, opened 2025-09-06
ProjectRed Framed Wire GT Recipes & PR Wire Refactor

UltraProdigy:
### Changes:
- Moved cable dye recipes to ScriptProjectRed instead of chemical bath recipes; makes more sense this way
- Got completely rid of the "ProjectRed Crafting" tab that's custom made by the mod and clutters NEI for no reason, replaced with our GT version
- Changed the recipes for all framed wire variants, they all used forge microblocks (ew, I know) before. Now they still have the same cost, but in sticks for crafting, and have an added packager recipe that is consistent with the already greggified wire recipes- not sure why these ones were looked over before.
- Added unpackager recipes, and changed a couple unpackager recipes that fabricated rubber sheets out of thin air
- Moved around code, and refactored it into much neater loop code for readability (which is why there are so many line deletions).

Tested in game, nothing is removed and nothing is changed besides the framed wire recipes and the addition of a couple packager / unpackager recipes that should have been there before for consistency's sake. This will not affect balance at all.

### Some Old Photos:
image
image

### Some New Photos:
image
image


EnderProyects (approved):
LGTM

PR #1366, opened 2025-09-06
Replace Pam's candle with torch in EFR lantern recipe

Kynake:
image

Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/21134


src/main/java/com/dreammaster/scripts/ScriptEFR.java
```diff
@@ -886,7 +886,7 @@ public void loadRecipes() {
                 getModItem(EtFuturumRequiem.ID, "lantern", 1, 0, missing),
                 bits,
                 new Object[] { "IGI", "PCP", "III", 'I', "plateIron", 'G', "dustGlowstone", 'P', "paneGlassColorless",
-                        'C', getModItem(PamsHarvestCraft.ID, "pamcandleDeco1", 1, 0, missing) });
+                        'C', getModItem(Minecraft.ID, "torch", 1, 0, missing) });
```
EnderProyects, line 889: Instead of Minecraft.ID use Minecraft own "ItemStack(Blocks.torch"  It should be something like that

EnderProyects (changes requested):

src/main/java/com/dreammaster/scripts/ScriptEFR.java
```diff
@@ -886,7 +886,7 @@ public void loadRecipes() {
                 getModItem(EtFuturumRequiem.ID, "lantern", 1, 0, missing),
                 bits,
                 new Object[] { "IGI", "PCP", "III", 'I', "plateIron", 'G', "dustGlowstone", 'P', "paneGlassColorless",
-                        'C', getModItem(Minecraft.ID, "torch", 1, 0, missing) });
+                        'C', new ItemStack(Blocks.torch) });
```
EnderProyects, line 889: You also need to add the amount and the meta, like ", 1, 0", I suppose it will work without that since it should be meta 0

Kynake: It works without. `new ItemStack(item)` will call the other constructor, default to size 1, metadata 0, which works for the torch

EnderProyects (approved):

PR #1372, opened 2025-09-13
Change charged battery to empty for treetap and hoe

Miklebe:
Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/21244
Changed the charged battery

image

to empty one
image


Miklebe:
I tried to make the recipe like that of windmeter
image

But a lot of IC2 tool recipes are also removed from the coremod, so couldn't track what made the recipes for them have oredict battery or even appear in the full pack (they are missing in coremod env)

Miklebe:
If there was any specific reason as to why some recipes are removed from loader and manually added in GT_CraftingRecipeLoader, i will revert last commit.

PR #1375, opened 2025-09-17
Hides Space Research and Space Project Modules in NEI

UltraProdigy:
Sister PR to https://github.com/GTNewHorizons/GT5-Unofficial/pull/4884
Tested in game


Miklebe:
Since you hid some stuff already, might hide Entropy Processor and it's casings? (i didn't do it cus was confused as to how to specify ids for blacklist, but i was thinking of editing config directly)

PR #1393, opened 2025-10-02
Disable Open To LAN

wlhlm:
There are many issues with Open To LAN in GTNH, mostly stemming from Betterquesting IIRC. To prevent players from accidentally trying it, disable the button with a tooltip with advice for using a dedicated server. Also, a mixin is provided to prevent starting the integrated server by other means.

I hope the tooltip is not too wordy.

The tooltip is on a small delay to prevent it from constantly popping up while just moving your cursor around in the menu.


boubou19:
Some people aren't going to like it, but it's better like this.

Dream-Master:
we force terraingen as well

Dream-Master:
> I wouldnt say "force". You can easily use something else like RTG if you disable DWG

Okay, we'll guide people in the right direction. Better?

Dream-Master:
> NOOOO, You cant turn world in creative from survival for testing without this...

no that work fine even if you not be in lan

Flakxo:
> > NOOOO, You cant turn world in creative from survival for testing without this...
> 
> no that work fine even if you not be in lan

what if the player didnt enable cheats when they created the world? for example when i use worlds downloaded from other people, i need to open to lan and enable cheats which allows me to interact with their claims

mitchej123:
I am overall in favor of this, but if people did not create a world with cheats enabled, this prevents them from enabling cheats.  There needs to be a workaround, or this needs to be a soft disable with shift/ctrl override.

mitchej123:
> > NOOOO, You cant turn world in creative from survival for testing without this...
> 
> no that work fine even if you not be in lan

It does not if you did not enable the world with cheats.  Switching to LAN & enabling cheats allows this.  There should be another alternative or this should only be a soft disable.

Caedis:
Rename the button to "Enable Cheats" and have it just enable cheats

Dream-Master:
> Rename the button to "Enable Cheats" and have it just enable cheats

no this will not resolve the problem people use this often and report various bugs

wlhlm:
I honestly forgot this was a thing. :caught:

My main concern here was to prevent bug reports from players playing in multiplayer via Open to LAN. As such, I don't think a soft block is enough.

I have checked the code, and it looks like for as long as no other player connects to the "open to LAN" session, it's effectively the same game/environment as a pure single player session, nothing really happens on the backend, only that you can't pause your game anymore. Here is all that happens (`net.minecraft.server.integrated.IntegratedServer#shareToLAN`):

1. open port so that other players can join
2. set session to public (prevents the game from pausing)
3. start discovery ping so that the server is found by other instances in the network
4. set gamemode and whether to allow cheats for the current session

I can just disable 1-3 via mixin and have it print a message to chat saying that nobody will be able to connect.

This effectively turns the "Share to LAN" button into the "Enable cheats" button as eluded to above, but it feels somehow weird doing that directly.

Does that sound fine?

PR #1397, opened 2025-10-07
downtier wormhole generator from UHV -> ZPM so it gets more use

chrombread:
as title says. discussed briefly in meta-dev https://discord.com/channels/181078474394566657/939305179524792340/1424977986519699456

this multi is really cool but its gated a little bit too late to really get use. this downtier doesn't really change much since it was using zpm/late-zpm casings anyway. it'll just be a cooler active transformer.

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/AssemblingLineRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/AssemblingLineRecipes.java
@@ -1625,24 +1625,24 @@ public class AssemblingLineRecipes implements Runnable {
                 (int) TierEU.RECIPE_UMV);
 
         // Miniature Wormhole Generator
-        TTRecipeAdder.addResearchableAssemblylineRecipe(
-                AEApi.instance().definitions().materials().singularity().maybeStack(1).get(),
-                64000,
-                64,
-                200000,
-                4,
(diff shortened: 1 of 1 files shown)


Dream-Master:
This are 3 tiers down so a huge buff

chrombread:
its currently more or less useless. this would tier it from mid-uhv to late-zpm. its main quirk is wireless power transfer but it gets overshadowed by uhv cap LSC's. thats why i think downtiering it here would make it used a lot more, since its quite a cool multi

Dream-Master:
what about put this at starting with uv so it is one tier earlier ?

chrombread:
it can be, but there isn't really much of a gap between late-zpm and early-uv i dont think. I could just swap one of the fluids for americium and call it a day, i guess.

chrombread:
Americium is just a fluid that is gated by the T2 fusion reactor, which is the barrier point to UV. as far as I know at least, its been a while lol

Dream-Master:
> @Dream-Master I can always add some materials from my line in here when me and noc are done with it ( since its teleport stuff), the end of the line is Uv thought, but some of the fluids are ZPM

if your pr was added first and after a discussion sure why not

Dream-Master (approved):

PR #1399, opened 2025-10-07
Fixed missing recipes

cubefury:
Fixes missing recipes for VM blocks. This just fixes the blocks being craftable for now - We're pending some changes to retier the block, which will be implemented in the coming days.


EnderProyects:
It could be nice to lock the vending machine behind coins ( Use a coin in the recipe)

Dream-Master (approved):

PR #1533, opened 2025-12-31
Ender-io QoL

EnderProyects:
Add recipes from "The Vat" from Ender-io in the BacVat, added them so it uses the max item for max fluid, also added a poisonous potato recipe in order for one to work well.

image

image
image
image
image
( here changed sunflower to sunny quartz, because sunflower changes to lilac for some reason idk)
image
image
image
image


EnderProyects:
Reduced the fluids from 1000 to 10 and increased the items from 1 to 4, also changed water used in some recipes to mineral water ( makes more sense for clouds and that stuff and also makes so you cant slap a resorvoir hatch in the vat)

UltraProdigy (commented):
Love the no sieverts. I think the dust cost in the Vapor of Levity should be halved, but if you're set on the current amount I'll cede since it's feasible

UltraProdigy (approved):
Yeah you're right it's fine. 
This looks good

PR #1540, opened 2026-01-03
Ender io qol part2 ( Soul Infused Medium Recipes)

EnderProyects:
Added new fluid and new recipes to create Ender-io items normally crafted in the Soul binder using soul vials, the new fluid is like a concentrated of soul material, and that reacts with the corresponging mob drop and binds to the item ( the new fluid/line requires to set up first soul binders, since it requieres vapor of levity, so it doesnt remove usage), also added a recipe to get the enchanted Totemic Capacitator, so you can speed up other Ender-io machines:

Depends on this in order to work: https://github.com/GTNewHorizons/GT5-Unofficial/pull/5630
And not needed but part of it: https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/1533

image

 Recipe for the Soul Infused Medium ( all soul vial recipes main loot, and some eassy to obtain fluid, nothing magic related)
 
 
image

Recipe for upgraded totemic capacitor

image
image
image
image
image
image

The ammount of Soul infused medium is 144 x xp levels requiered for original recipe


EnderProyects:
Well for the items the idea is that you obtain those via EEC since they are mobdrops or related, so: soularium for the subtitution of soul vials, emerald dust for villagers souls, enderman heads for enderman souls, zombie brain for zombie souls, and witches hand for witches soul, ( all the 4 souls used in the recipes), and then a heart.
- For the item part I could remove the soularium and the heart, and reduce the item count to only 1 of each

The fluids are eassy obtainable in Ev, experience, blood, cadaverine and putrefactine ( unused fluids extracted from the death, I think are nice to give them a use), and then grow medium and vapor of levity, to give a loop in the recipe and is just not a better recipe over the normal one.
- For the fluid part I could remove both putrefactine and cadaverine, and also the growth medium, leaving experience, blood and vapor of levity

UltraProdigy (commented):
So I love the idea of being able to skip the EnderIO machines, they are largely manual which is quite annoying. The problem is that these recipes are so much work to do that a reasonable player is almost never going to prefer this to just hitting the manual XP button. I'm commenting here but this encompasses your part 1 as well. The Bacterial Vat is a pain in the butt to setup and regulate, this is essentially asking for at least 3 dedicated bac vats & their bacteria cultures, and input ramping, and radiation providing, just to get the medium, all to make things that are used very infrequently and aren't needed for completing the game. Also, the recipes are recursive which adds another layer of headache for the player.. something I would only expect out of a required (or optional but beneficial) line.

I think this needs to be re-engineered to be either more rewarding for the player to setup, or far less difficult.

And this is mostly regarding the creation of soul infused medium, the QOL autoclave recipes are fine.

EnderProyects:
> Also, the recipes are recursive which adds another layer of headache for the player.. something I would only expect out of a required (or optional but beneficial) line

If done correctly it doesnt need loops, also as I said early in another coment I could simplify the recipe

LazyFleshWasTaken:
Imo, three different fluids for 2-ish items each, i.e., zombie./skeleton, villager/witch, enderman may be better? It would simplify the recipe, and would take out the emerald and witch hand from the more common use cases of enderio (spawner/eec) 

Also, didn't the totemic capacitor get booted from the stargate recipe?

EnderProyects:
About the last part, I don't know why totemic is related in here

EnderProyects:
The replacement for totemic capacitators use one of the crystals that takes enderio soul bonding

UltraProdigy (commented):
What if we equalized the amount of putrescine & cadaverine & got rid of the red heart in the growth medium recipe?

UltraProdigy (approved):

PR #1566, opened 2026-01-17
Necessary changes to adapt to biolab cleanup

boubou19:
Adapt cultures to work after https://github.com/GTNewHorizons/GT5-Unofficial/pull/6892


src/main/java/com/dreammaster/bartworksHandler/BacteriaRegistry.java
```diff
@@ -59,120 +59,46 @@ public void runAllPostinit() {
...
-        );
-        BioCulture bioCulture = createAndRegisterBioCulture(
-                new Color(133, 0, 128), // color = violet
-                "Barnadafis Arboriatoris", // name
-                BioPlasmid.convertDataToPlasmid(bioData), // BioData -> plasmid
-                BioDNA.convertDataToDNA(bioData), // /BioData -> DNA
-                EnumRarity.rare, // rare (only visual)
-                true // can be multiplied in the BioVat
-        );
+        BioCulture bioCulture = BioCultureEnum.BarnadafisArboriatoris.bioCulture;
```
Guvante, line 62: Should these just be inlined?

boubou19: definitely

src/main/java/com/dreammaster/bartworksHandler/BacteriaRegistry.java
```diff
@@ -98,136 +87,12 @@
 
 public class BacteriaRegistry {
 
-    static final LinkedHashMap<String, BioCulture> CultureSet = new LinkedHashMap<>();
-
     public void runAllPostinit() {
-        runLateBioOBJs();
+        regenerateBioFluids(); // this will generate bacteria fluids. needs to be called AFTER ALL breedable bacterias
```
Guvante, line 91: Niptick: A comment explaining why that has happened would make this clearer (since you can't tell from here that it was done correctly)

Guvante (approved):
I don't know about the gradle changes but I don't see any typos in the registry migration.

Unrelated to this change which is fine, I think renaming the enum to the fancy name makes the code harder to read, I think leaving the old strings as the code name would have made these migrations cleaner.

PR #1688, opened 2026-04-07
Rebalance Recipes for Super Chest/Tanks 2 & 3

NotKaizul:
Currently Super Tank 3 is craftable before Super Tank 2, as 2 reqires a vacuum freezer to process vibrant alloy.
Energetic Silver makes more sense as the plate material for the recipe, since it only requires Kanthal coils.


UltraProdigy:
@NotKaizul Run ./gradlew spotlessApply & commit

Dream-Master (changes requested):
I just put this on hold . It need further discussion.

LazyFleshWasTaken:
>  The Vibrant would be for gating it past freezer.

That would put it at tier + 2. Lv field gen is already almost hv, since it needs hv circuits, which are now the last gate to hv. With energetic, its only after knathal coils, whcih are firmly mv now.

NotKaizul:
The mv, hv, ev, and iv field gens use titanium, niobium-titanium, hss-g, and hss-s respectively. None of these have anywhere near a full tier of progression between them, so I don't see an issue with making lv field gen use vibrant alloy. Even energetic alloy would put the lv field gen quite far before the mv one.

UltraProdigy (dismissed):

LazyFlesh (approved):
lgtm

UltraProdigy:
Superseded by https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/1700

PR #1689, opened 2026-04-08
Add a Machine Recipe for Coated Circuit Boards

NotKaizul:
Coated circuit boards are used for 1k ae2 cells, so it would useful to have a machine recipe for them. 
Was mentioned in [#beta-testing](https://discord.com/channels/181078474394566657/522098956491030558/1491149903206944882).
Currently the recipe cost is identical to the crafting table recipe. I think the assembler recipe should be a bit cheaper to motivate players to use it pre-ae2, but there should be some more discussion about what that would look like first.
Here's the recipe (both have the same eu/t and time)
image

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
@@ -3753,17 +3753,6 @@ public class AssemblerRecipes implements Runnable {
 
     private void makeCircuitPartRecipes() {
         // Circuits and Boards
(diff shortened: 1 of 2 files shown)


NotKaizul:
great idea ima do that

LazyFleshWasTaken:
I'd rather have the recipes be multiplied, see ones for full board: 
image

just take out copper and you have em, keep em in chembath if you want.

NotKaizul:
right. should be fine now since its ina different machine

LazyFlesh (approved):

PR #1705, opened 2026-04-14
Change plasmaArcFurnaceRecipes to normal maps

kuba6000:
- Changed plasma recipes to normal arc furnace map
- Changed chronic singularity recipe to be UMV instead of UXV because arc furnace lost the ability to tier skip

Merge after https://github.com/GTNewHorizons/GT5-Unofficial/pull/6342


src/main/java/com/dreammaster/scripts/ScriptGalacticraft.java
```diff
@@ -1340,357 +1339,290 @@ private void arcFurnaceRecipes() {
...
                 .itemOutputs(
                         ItemList.Ingot_Heavy1.get(11L),
                         GTOreDictUnificator.get(OrePrefixes.ingot, Materials.MeteoricIron, 5L),
                         GTOreDictUnificator.get(OrePrefixes.ingot, Materials.Steel, 42L),
                         GTOreDictUnificator.get(OrePrefixes.ingot, Materials.StainlessSteel, 4L))
-                .outputChances(10000, 10000, 10000, 10000).fluidInputs(FluidRegistry.getFluidStack("oxygen", 6000))
-                .duration(1 * MINUTES).eut(TierEU.RECIPE_LV).recipeCategory(RecipeCategories.arcFurnaceRecycling)
-                .addTo(arcFurnaceRecipes);
+
+                .duration(1 * MINUTES).eut(TierEU.RECIPE_LV).metadata(RECYCLE, true).addTo(UniversalArcFurnace);
```
danyadev, line 1624: noooooo while replacing the lines an empty line has sneaked in..........

danyadev: unfortunately there are much more than one empty line (because it was most probably a find and replace operation, so an empty line was added in many places at once)

UltraProdigy:
Conflicts

src/main/java/com/dreammaster/scripts/ScriptGalacticraft.java
```diff
@@ -1898,758 +1829,6 @@ private void maceratorRecipes() {
                 .duration(15 * SECONDS).eut(2).addTo(maceratorRecipes);
     }
 
-    private void plasmaArcFurnaceRecipes() {
```
Guvante, line 1901: Looking at the Arc Furnace PR it appears that adding the tag RECYCLE skips Arc Furnace recipes that use plasma (aka the new version of Plasma Arc Furnace recipes)

https://github.com/GTNewHorizons/GT5-Unofficial/pull/6342/changes#diff-6368dbef138be0be79cfdbb5fb1d7a04e23d3d053b645d6be45a0c0a90a56b3eR339

But then this PR has quite a few recipes that had Plasma Arc Furnace recipes while notating the equivalent Arc Furnace recipes (which to be clear also had Arc Furnace recipes) as Recycle meaning these Plasma Arc Furnace recipes won't exist anymore in the future.

I don't have opinions on this from a balance standpoint, only pointing out the difference to ensure it was intentional.

Guvante (commented):
Newline cleanup would be nice but I didn't dive deep into this since the only things that caught my eye were balance related

src/main/java/com/dreammaster/scripts/ScriptGalacticraft.java
kuba6000: Doesn't really matter, I don't think anyone used these recipes anyway

OTPANNIEXD (approved):
lgtm

PR #1706, opened 2026-04-15
Add sulfur block

EnderProyects:
Add sulfur block to be used in the crops mod and well, anywhere
image


LazyFleshWasTaken:
shouldn't this be added to gt5's sulfur's material builder, as the block, a la glowstone/redstone?

EnderProyects:
Wont that create a base texture instead of a custom one? ( Also idk why is failling in the compression)

C0bra5 (changes requested):
Crops NH shouldn't be the thing to add a recipe for this block; it should preferably be declared in the core mod (a.k.a where the block is declared) or in GT5u, preferably via the material system.

C0bra5 (approved):

PR #1725, opened 2026-04-24
don't load the draconic evolution earth recipe if the block is not registered

Alexdoru:


src/main/java/com/dreammaster/scripts/ScriptDraconicEvolution.java
```diff
@@ -1235,16 +1236,24 @@ public void loadRecipes() {
...
+                            getModItem(ExtraUtilities.ID, "greenscreen", 64, 13),
+                            getModItem(OpenComputers.ID, "hologram2", 1, 0),
+                            createItemStack(
+                                    ExtraUtilities.ID,
+                                    "microblocks",
+                                    1,
+                                    3,
+                                    "{mat:\"tile.extrautils:decorativeBlock1_5\"}"))
+                    .fluidInputs(new FluidStack(FluidRegistry.getFluid("water"), 10000))
+                    .itemOutputs(getModItem(DraconicEvolution.ID, "earth", 1, 0)).duration(6 * SECONDS).eut(200)
```
boubou19, line 1254: the GMI call is already done here, so you probably can reuse it and cache it to avoid one lookup

Alexdoru: It's ok we're not gonna die from one more lookup

boubou19: still odd to do it differently, but does the job

boubou19 (approved):

PR #1726, opened 2026-04-24
Fix crash when some gt ore stonetypes are disabled

### LINKED ISSUE #22965: Freeze at the end of startup when some GT stone generation settings are disabled

aloaloolaola

### Your Pack Version

2.8.3

### Your Server

N/A

### Java Version

Java 25

### Type of Server

None

### Your Actions

Fresh prismlauncher 2.8.3 instance from the prism zip download, no configs or anything changed.
starts up fine

Disable the generation of a stone type in .minecraft/config/GregTech/WorldGeneration.cfg (tested with B:generateBasaltOres=false)
start the game again
The startup process goes normally until at some point, seemingly always during "Post Initialization: loading GregTech" at around 80-ish percent, the startup complete ping sound plays, the window becomes unresponsive and cpu usage drops to 0%. Memory usage is reasonable and not the issue.

From the crash report, this seems to be due to some recipes failing to be created, but i would say that the game freezing instead of crashing cleanly is also a problem itself.

This happens on java 25 and java 21 from different distributors.

### Crash Report

[crash-2025-12-22_02.01.47-client.txt](https://github.com/user-attachments/files/24282118/crash-2025-12-22_02.01.47-client.txt)

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the crash still exists will prompt us to investigate and reopen it once we confirm your report.
- [x] I can reproduce this crash consistently by following the exact steps I described above.
- [ ] I have asked other people and they confirm they also crash by following the exact steps I described above.

tiffit:
Sibling PR: https://github.com/GTNewHorizons/GT5-Unofficial/pull/6492

Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/22965


src/main/java/com/dreammaster/tinkersConstruct/MeltingRecipeAdder.java
```diff
@@ -23,12 +23,16 @@ public class MeltingRecipeAdder {
...
         this.meltingTemperature = meltingTemperature;
         this.fluidName = fluidName;
-        addMelting = itemStack -> Smeltery.addMelting(
-                itemStack,
-                renderBlock,
-                renderBlockMeta,
-                meltingTemperature,
-                FluidRegistry.getFluidStack(fluidName, amount));
+        addMelting = itemStack -> {
+            if (itemStack != null) {
```
Ruling-0: Is this null guard necessary? Since you don't add the prefixes, I'd expect this not to add to the fix. If it's not possible for this to _intentionally_ encounter null, we should leave it unguarded so it properly errors when a null is unintentionally present.

tiffit: Its there because all of the ores are added as one big array to the smeltery https://github.com/GTNewHorizons/NewHorizonsCoreMod/blob/master/src/main/java/com/dreammaster/scripts/ScriptTinkersConstruct.java#L1697 so adding a null check here was the easiest. I could also rewrite that chunk of code so that it checks for if the stonetype is enabled.

Ruling-0: I think it's better practice to just rewrite the other chunk of code.

tiffit:
Rewrote how ores are added to the smeltery and it should be a lot easier in the future to add ores and stone variants as well

src/main/java/com/dreammaster/scripts/ScriptTinkersConstruct.java
```diff
@@ -1690,115 +1692,52 @@ public void loadRecipes() {
...
-                GTOreDictUnificator.get(OrePrefixes.ore, Materials.GraniticMineralSand, 1L),
-                GTOreDictUnificator.get(OrePrefixes.rawOre, Materials.GraniticMineralSand, 1L),
-                GTOreDictUnificator.get(OrePrefixes.oreNetherrack, Materials.GraniticMineralSand, 1L),
-                GTOreDictUnificator.get(OrePrefixes.oreEndstone, Materials.GraniticMineralSand, 1L),
-                GTOreDictUnificator.get(OrePrefixes.oreBlackgranite, Materials.GraniticMineralSand, 1L),
-                GTOreDictUnificator.get(OrePrefixes.oreRedgranite, Materials.GraniticMineralSand, 1L),
-                GTOreDictUnificator.get(OrePrefixes.ore, Materials.Magnetite, 1L),
-                GTOreDictUnificator.get(OrePrefixes.rawOre, Materials.Magnetite, 1L),
-                GTOreDictUnificator.get(OrePrefixes.oreNetherrack, Materials.Magnetite, 1L),
-                new ItemStack(GregTechAPI.sBlockOres1, 1, 28706),
```
Ruling-0, line 1727: Can you confirm in-game that these manual calls (I believe this is end magnetite ore) were correctly translated to the new system?

tiffit: Can confirm they translate properly

Ruling-0 (approved):
Very nice work and cleanup!

PR #1738, opened 2026-05-06
Add recipe for Potion of Blissful Oblivion

### LINKED ISSUE #24272: Add a "Potion of Blissful Oblivion" as a method of salvaging run for inexperienced players

AwonGardino

### Your Pack Version

2..8.4

### Your Proposal

My proposal is to give new, less experienced players, a last ditch attempt to save their run and solve their warp problems once and for all in a small price of complete deletion of their thaumcraft progress.

The Potion of Blissful Oblivion
Image

Relatively cheap recipe for a potion, that will remove all warp at once, but delete all thaumcraft progress from user data. Every research, every scan, every piece of knowledge is gone for good. If player wants to do it again - they would need to start from ground 0. Well, except the stuff they already crafted.

Additionally, we could make a quest entry, that will tell players that they can craft it. Maybe, some triggers to spotlight it like "Warp over 100 + /iskamiunlocked less than 75% complete", etc.

Just in case if player does not understanding what are they doing, make this potion very slow to consume and leave a message in chat that clearly says, that this will delete all progress.

### Your Goal

The goal is simple - to not lose new players, that poured tens of hours into modpack and "entered a wrong door", therefore ruined themselves with very high amounts of warp. I saw this too many times. Either they grab the one ring, researched elemental fires, or just consumed every taboo research they could reach without realizing the consequences or severity of their actions. 

### Your Vision

In my playthroughs I met too much new players, that got overwhelmed by amount of researches in thaumcraft, gained outstanding amount of warp at LV-MV and, basically, ruined their run. While, there is numerous way of warp control and containment even at low levels, those players are getting overwhelmed, extremely upset and leaving modpack without any attempt to solve the issue, even if there is a lot of people to help them out. But, i think, if there would any way to salvage their run in cost of thaumcraft progression, that could have saved us some of those players.

Yes, I understand that this could be used for cheesing, but I truly believe that this addition will make new players' experience a little bit better.

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this feature please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [x] I believe there is nothing similar in the pack already, or the existing solution isn't good enough.
- [x] I understand this change request may not attract enough attention and thus not be implemented.
- [x] I understand this change request may be rejected due to other community members thinking it's inappropriate.
- [x] I believe this feature would make the pack better.

koolkrafter5:
image

Also set the potion's aspects to Strontio and set the aspects for Hunk of Something to include Vesania (insanity) for some fitting sources of these rather rare aspects.

image

image

Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/24272

### DIFF
--- a/src/main/java/com/dreammaster/scripts/ScriptWarpTheory.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptWarpTheory.java
@@ -263,5 +263,8 @@ public class ScriptWarpTheory implements IScriptLoader {
         ThaumcraftApi.registerObjectTag(
                 getModItem(WarpTheory.ID, "item.warptheory.oblivionpotion", 1, 0),
                 new AspectList().add(TCAspects.STRONTIO.getAspect(), 8));
+        ThaumcraftApi.registerObjectTag(
+                getModItem(WarpTheory.ID, "item.warptheory.something", 1, 0),
+                new AspectList().add(TCAspects.VESANIA.getAspect(), 2).add(Aspect.HUNGER, 4).add(Aspect.FLESH, 4));
     }
 }
--- a/src/main/java/com/dreammaster/scripts/ScriptWarpTheory.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptWarpTheory.java
@@ -17,10 +17,14 @@ import static gregtech.api.enums.Mods.Witchery;
 import java.util.Arrays;
 import java.util.List;
 
+import net.minecraft.init.Items;
+import net.minecraft.item.ItemStack;
+
 import com.dreammaster.thaumcraft.TCHelper;
 
 import gregtech.api.enums.Materials;
 import gregtech.api.enums.OrePrefixes;
+import gregtech.api.enums.TCAspects;
 import thaumcraft.api.ThaumcraftApi;
 import thaumcraft.api.aspects.Aspect;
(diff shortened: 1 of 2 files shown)


koolkrafter5:
Hmm, maybe using the erasers in the recipe isn't the best idea. They leave behind damaged versions of themselves which you can craft together to get more erasers using the regular tool combination recipe.

https://github.com/user-attachments/assets/cb4d4991-554c-4353-9c5b-2a131976d113

koolkrafter5:
On the other hand, it is kind of funny. I'll leave it up to a reviewer if I should change it or not.

EnderProyects:
I would change the erasers to arcane litmus paper, and the watter bottle to a bottle o´ enchanting, but this recipe seems fine, apart from the eraser problem

Dream-Master:
> Hmm, maybe using the erasers in the recipe isn't the best idea. They leave behind damaged versions of themselves which you can craft together to get more erasers using the regular tool combination recipe.
> 
>  erasereraser.mp4

why is this a bad idea. Did something can be duped ? @EnderProyects Litmus paper looks a bit wired and make no sense for me.

Dream-Master (approved):

PR #1740, opened 2026-05-07
Use OreDict for Shuttle Schematic

czqwq:


czqwq:
the replace was not successful,i'm still testing

czqwq:
image

working fine now
tested in prod envs

tiffit:
Should we allow raw ores as well?

czqwq:
> Should we allow raw ores as well?我们是否也允许使用原始矿石？

it just match all ores but no rawOre

I think this was deliberately designed.

Nana-Sakura (approved):

Nana-Sakura:
> Should we allow raw ores as well?

absolutely not

LazyFlesh:
> > Should we allow raw ores as well?
> 
> absolutely not

Why? If you have access to these ores, you can get an infinite supply. Sure, you can fortune raw ore, but either people slap down a VM and have infinite, or they go around fortuning what they need. I don't see any balance concerns to not have em.

czqwq:
My PR is only responsible for implementing the acceptance of variant ores for minerals. As for whether to add raw ores, please open another PR for discussion.

Yoshy2002:
> My PR is only responsible for implementing the acceptance of variant ores for minerals. As for whether to add raw ores, please open another PR for discussion.

It was discussed in meta dev, you implemented it and then removed it again without any new discussion if i may add this here.

czqwq:
> > My PR is only responsible for implementing the acceptance of variant ores for minerals. As for whether to add raw ores, please open another PR for discussion.
> 
> It was discussed in meta dev, you implemented it and then removed it again without any new discussion if i may add this here.

If most devs agree to add it after discussion, you can manually submit a PR and link it to the current PR. Anyway, I don't really understand why we need to add rawOre

PR #1748, opened 2026-05-10
Replace Aspect.getAspect with direct references to aspect objects for thaum aspects

koolkrafter5:
Aspect.getAspect("tag") needs to search through a LinkedHashMap to find the aspect. Why do that when you can just directly access them? This PR is a simple find and replace to do that.

Adds Forbidden Magic and Magic Bees as deps for direct access to their aspects.

Draft for now because I found a material registration that still uses Aspect.getAspect() while writing this.


UltraProdigy (approved):
Dep conflict needs to be resolved but otherwise looks good and works well 👍

PR #1750, opened 2026-05-10
Make sulfur from cloud recipe less restrictive (3x3 into 2x2)

Algent:
On @tiffit request

Tested ingame (full pack):
image


tiffit (approved):
:)

EnderProyects:
There is a recipe to get the sulfur cloud ( I don't remember the recipe) did you check that is not sulfur positive?

tiffit:
> There is a recipe to get the sulfur cloud ( I don't remember the recipe) did you check that is not sulfur positive?

This PR doesn't change the recipe. Just makes it so the 2x2 clouds don't need to be in the top left

EnderProyects:
Ah I misunderstand then 😅

PR #1751, opened 2026-05-11
Fix Quick and Dirty Portal Gen recipe

### LINKED ISSUE #24101: Dirty portal generator assembler recipe bugged

ableytner

### Your Pack Version

daily-440

### Your Server

private server

### Java Version

Java 25

### Type of Server

None

### Your Expectation

Crafting a dirty portal generator with an industrial diamond and 4 LV circuits in an assembling machine should craft a dirty portal generator, as seen in NEI.

Image

### The Reality

The assembling machine doesn't start the craft. If I use a vanilla diamond instead, the recipe works as expected.

Strangely enough, if it was crafted once, the industrial diamond recipe works as expected in that specific assembling machine. (Seems like the correct recipe gets cached then)

This problem is consistently reproducible in NewHorizonsCoreMod dev environment.

### Your Proposal

fix

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [x] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [ ] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

tiffit:
Fixes industrial diamonds not working to make the quick and dirty portal gen. I suspect its because the industrial diamond has 2 oredicts and gemDiamond is not the first one but I have no idea.

Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/24101

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
@@ -912,7 +912,7 @@ public class AssemblerRecipes implements Runnable {
 
         GTValues.RA.stdBuilder()
                 .itemInputs(
-                        GTOreDictUnificator.get(OrePrefixes.gem, Materials.Diamond, 1L),
+                        new OreDictItemStack(OrePrefixes.gem.get(Materials.Diamond).toString(), 1),
                         GTOreDictUnificator.get(OrePrefixes.circuit, Materials.LV, 4L))
                 .itemOutputs(NHItemList.TwilightCrystal.get()).duration(30 * SECONDS).eut(TierEU.RECIPE_LV / 2)
                 .addTo(assemblerRecipes);


UltraProdigy (approved):

Guvante:
Does the orignal call actually trigger an OreDict? I can't figure out why it would. (Your PR is wonderful, I am wondering how pervasive this bug is...)

tiffit:
> Does the orignal call actually trigger an OreDict? I can't figure out why it would. (Your PR is wonderful, I am wondering how pervasive this bug is...)

I'm not actually sure how recipe lookup works. The previous call would just return an itemstack using the given OrePrefix and Material. I assume something is failing during lookup when it looks for items with similar oreDict entries. This PR explicitly states the oreDict string which makes it so the lookup doesn't need to make any assumptions.

Guvante:
When I traced the bugged line I saw a bunch of replacements like to similar to the old code

But looking at the code it appears that the old format just generated a stereotypical Item stack which then gets stored as a singleton...

But then again it obviously eventually transforms to an OreDict so I am missing something.

PR #1759, opened 2026-05-13
Add Circuit to Jabba Barrel Recipe

NotKaizul:
Currently has a conflict with efr barrel recipe
now fixed:
image


Caedis:
Does the EFR barrel have a circuit?

NotKaizul:
idk why i thought circuit recipes take priority over non-circuit recipes
> Does the EFR barrel have a circuit?

it does now!!!

Caedis:
why circuit 13 and 14?

NotKaizul:
the jabba structural upgrades also use circuit 13
14 cause its next to 13 idk

tiffit:
13 and 14 does feel a little weird, would be nice if it was more commonly used ones

UltraProdigy:
@NotKaizul Let's move Jabba recipe to C1 and EFR barrel to C2. This PR is held up bc using 13 & 14 is inconvenient. 

As for the JABBA upgrades I've only looked at a couple of them so this needs to be double checked for conflicts but they can potentially move to C1 as well in a separate PR if we think that's warrented

UltraProdigy (approved):
awesome

PR #1760, opened 2026-05-14
Xenoxene culture recipe in Beamcrafter after initial xenoxene

ham-corp:
image

Added recipe to make xenoxene xenoxsis culture, after getting access to xenoxene.

Making multiple cultures the biolab way kinda sucks, and my "lore" here is that we're irradiating the culture to mutate it.

Also the texture colours happen to line up nicely, and xenoxene xenoxsis is still barnarda + tau (particles instead of cetiE culture🤓)


Auynonymous (approved):
Looks good to me, Perhaps having the culture at a 100% chance output might be a bit too easy. Maybe a lower rate? 50%?

Dream-Master:
> Looks good to me, Perhaps having the culture at a 100% chance output might be a bit too easy. Maybe a lower rate? 50%?

@ham-corp thoughts ?

ham-corp:
> > Looks good to me, Perhaps having the culture at a 100% chance output might be a bit too easy. Maybe a lower rate? 50%?
> 
> @ham-corp thoughts ?

Yeah seems fair, I'll change it

PR #1761, opened 2026-05-15
Adds missing singularities to BHC

OTPANNIEXD:
Adds missing singularities for which fluid is space pumpable to BHC.


DylanTaylor1 (approved):
Completeness :)

PR #1764, opened 2026-05-17
Add Assembler Recipes for Volcanus and Cryogenic Freezer Controllers

Auynonymous:
title lol

recipes are 10 seconds iv

no pictures cause gtpp mats are fire for some reason? but its the same cost obv

These controllers are gaining recipes in assembler due to the new MVF and MEBF using them in recipes, and will most likely be used more often in endgame because of this (like how dtpf uses mebfs)


chrombread (approved):
thanks

PR #1766, opened 2026-05-22
Pushback Pattern Repeater to UV Voltage

UltraProdigy:
Talked about here onwards https://discord.com/channels/181078474394566657/939305179524792340/1503323423051616357

Nerfed to UV Voltage & 256x Co-Processing Unit


GDCloudstrike:
what if the recipe also used a higher capacity co-processor? like the 1024x one for example
since currently its just ev parts in a uv recipe, which is a bit weird

UltraProdigy:
Great idea, I can switch it over.. let me just ponder costs real quick

boubou19 (approved):

PR #1769, opened 2026-05-25
Add a replacement recipe for creative(Magic) memory

czqwq:
add a replacement recipe for creative memory
image

### DIFF
--- a/src/main/java/com/dreammaster/scripts/ScriptComputronics.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptComputronics.java
@@ -55,7 +55,7 @@ public class ScriptComputronics implements IScriptLoader {
                 "ingotInfinity",
                 'i',
                 getModItem(Avaritia.ID, "big_pearl", 1));
-        // second replace recipe(big pearl->UIV field generator)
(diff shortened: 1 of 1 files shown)


EnderProyects:
Why?

EnderProyects:
- Magical Memory
- Removed the magic from the recipe...

czqwq:
> * Magical Memory  魔法记忆
> * Removed the magic from the recipe...把配方里的魔法都去掉了......

idk why this call magical memory,but we all call it creative memory

i just hope to add a recipe to skip the force play of the magic

Isn't it a good idea to replace non-costume big-pearl with a consume UIV field?

EnderProyects:
The problem is that we voted to keep magic mandatory and to integrate it more in the pack, not to remove it from recipes

czqwq:
> mandatory

i just add a replacement recipe,not remove them,it's optional

czqwq:
> > mandatory
> 
> i just add a replacement recipe,not remove them,it's optional

just like you add a iv field recipe to craft dim marker ,the secondly replacement recipe of the botania item

PR #1770, opened 2026-05-26
Adds End powder output to Ender goo recipe

OTPANNIEXD:
As in title, added at 7% chance.


GDCloudstrike:
I think this is unnecessary and just further trivializes things for no good reason. You already get enough dust from SE, and this recipe would make space mining end powder worthless by comparison.

EnderProyects:
I think the same, if you have problems with the space elevator just do the meteor ( that gives a lot) or the bee

Dream-Master:
i wondering where this 7% chance comes from?

UltraProdigy:
7% is arbitrary, it could stand to be much lower. I don't think this is problematic so long as it remains notably less efficient than mining in terms of EU. You'd still need an SE to pump fluid and a Spinmatron to keep pace which seems fair in theory.

Let me get in front of the game and do actual math first though. It's possible this isn't worth implementing but I'd like to check the numbers

UltraProdigy:
Did way too much analysis on this trivial issue. My conclusion: this isn't necessary. If anything were to happen we can just buff the SE output, but I don't think that's even necessary from my experiments. L suggestion on my part haha.

PR #1778, opened 2026-05-27
Add Styrene-Butadiene Rubber variants for Thermal Padding assembler recipes

luoyangyuli2001:
Adds Styrene-Butadiene Rubber as an alternative fluid input for all four Thermal Padding assembler recipes (Helmet, Chestplate, Leggings, Boots).

#1567 added an SBR variant for the Thermal Cloth recipe, but the Thermal Padding recipes were not updated accordingly. Currently Thermal Cloth can be made with either Silicone Rubber or SBR, while Thermal Padding only accepts Silicone Rubber. This PR fixes the inconsistency.

Disassembly recipes are unchanged as they do not involve rubber input.

---

I noticed this while playing GTNH 2.9.0-EXP-129. Thermal Cloth has an SBR recipe variant, but when I tried to craft the Thermal Padding armor pieces with SBR, there was no matching recipe. I've always been a Silicone Rubber enthusiast, but this time I wanted to try SBR based on suggestions from other players, only to find it wasn't possible at the padding stage. After checking past PRs, I found #1567 added the SBR variant for Thermal Cloth, so the omission of Thermal Padding variants seems like an oversight.

I attempted to test locally by replacing the jar in a full GTNH installation, but ran into version mismatch issues preventing world creation. That said, the changes are straightforward: each of the four new recipes is an exact copy of its Silicone Rubber counterpart with only the fluid input changed to `molten.styrenebutadienerubber`, following the same pattern already established in #1567 for Thermal Cloth.


UltraProdigy (approved):

UltraProdigy:
Ty for the PR :)

PR #1782, opened 2026-05-30
Add recipes for the Vending Machine Casing

YannickMG:
Actual casing texture subject to change.

https://github.com/GTNewHorizons/VendingMachine/pull/135 must be merged before this PR can be undrafted with an updated VendingMachine dependency version.

image

image


Dream-Master (approved):
I am ok with the casing changed from tin to iron. Maybe we can change the controller to be more expensive and the first one you get by wurst do you need only to craft the machine blocks

Ethryan (dismissed):
Nevermind, its not yet published as a version. Ignore this.

Ethryan:
Now it works.

PR #1786, opened 2026-06-01
Alternative Frenzy Frame recipe

1328116700:
QQ截图20260601125016

This recipe may look like it skips part of the bee progression, but it actually doesn't.
Since loot bags can be enchanted with Royal Jelly, there's a high chance of obtaining Magic Frames.
So, this recipe is just intended for players who have been unlucky.

https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/24815


LazyFlesh:
While I don't particularly like the recipe for some reason (I can't quite tell what), it doesn't seem like a _bad_ one. 

The mutagenic frame needs plut 241, which is really hard to get in fair quantities before you can get 239 ore, or 241 directly, since it's a 20% chance for a tiny dust. Presumably, 45ish 239 for 1 ingot 241, so about 2.5 stacks of 239 per mutagenic frame, or otherwise far in BM, and/or t6 planet.
Or got enough from 239 or even just uranium to breed trollplant, which also makes it (needs small 241 dust, an ev breeder, as well as the uranium crop, whcih further needs thorium and titanium crops)

Needing royal jelly also gates it to basically the same time as alveary, which you can add mutation and death chance to for basically free using electric stims, or an industrial apiary you can accelerate to hell, and add lifespan upgrades to. 

However, I would rather it use both royal jelly and pollen cluster. If it's going to be less rng (even more so than enchanting the bags with jelly), then it should require more upfront work, and tech tiering (1 tier, since it's possible in hv, reasonable in ev, while BM is hv) only bridges a part of that. If enchanting the bags increases the drop chance enough that an alt. recipe using it's enchant item is reasonable, then the player can choose what odds they want to gamble with (listed below)

Another option for an alt recipe is to use an output from bees that are on the way to getting one with soul comb, but earlier in the line (i.e. Arcane bee's enchanting drop. Arcane is 2-3 breeds away from spirit, which gives soul, and has a unique drop. Occult or otherworldy comb also works, and is a few more breeds down). This goes for frenzy frame as a whole, or as an alt recipe for magic frame (with appropriate wand tier increase, for using a less "potent" or "appropriate" focus, such as 150 vis to craft, or an infusion instead)

If you can make large amounts of mutagenic frames, you don't need this recipe anyway. Low durability aside, they have basically the same effect. Split its use with an oblivion frame, and you're fine. A strong WA would also do it.

Also, as an extra note: frenzy frames were added as a buff to bee breeding already, and as a reward for doing blood magic as well. Before, the best frame you could get was oblivion, or an elec stimmed alveary (tho you can't WA that). 

 
Overall, this leaves five options for the player if its added:
1. Get magic frames (or soulful wax) from unenchanted adv. bee bags (skips a lot of breeding, aside from the ones required to get the bags)
2. Get them from enchanted adv. bee bags, after breeding up to imperial bee (skips some breeding)
3. Make frenzy frames directly using this recipe after beeding up to imperial and industrious (if pollen is added to recipe), and being able to source plutonium 241 (ev to luv sourceable, with crop/bee -> meteor -> tech). (extra vis cost and using t4 instead of t3 slates is negligible) (doesn't skip much breeding, and you could honestly just go for spirit, vazbee, or w/e produces soul combs anyway)
4. Breed up spirit bee (most accessible producer of soul comb)
5. Skip using frenzy frames, and just use mutagenic frames. They have the same effect anyway, just a lower durability. More than enough to get you to a soul comb producing bee, especially when paired with oblivion.
 
 
 
 
Frankly, with how bees are, maybe removing soulful wax and the magic frames from the bags isn't out of the question, but I would leave such a determination up to dream/boubou.

LazyFlesh (changes requested):
see conversation.

Tldr: add pollen cluster to recipe, or change recipe for magic frame to use enchanting drop  and increase it's cost for using a "less potent/appropriate magical material".

Other changes would need more discussion.

1328116700:
> see conversation.
> 
> Tldr: add pollen cluster to recipe, or change recipe for magic frame to use enchanting drop and increase it's cost for using a "less potent/appropriate magical material".
> 
> Other changes would need more discussion.

Thanks for the suggestion.

This recipe is a less rng version of option 2, so the other ingredients aren't harder to obtain.
The Genetics questline gives a lot of loot bags, so there's roughly an 80% chance of getting Magic Frame/Necrotic frame/soul comb. In multiplayer, it's practically 100%. And those machines are basically useless.
At a 20% drop chance, it feels like a bonus for those who get lucky. At 80%, though, it ends up feeling more like a punishment for the players who don't.
So, I don't think it should be more difficult than option 2. In other words, I don't think Pollen Clusters should be added to the recipe.

This is assuming options 1 and 2 viable. If boubou/dream want to nerf loot bags, then this recipe wouldn't make sense anymore.

1328116700:
> see conversation.
> 
> Tldr: add pollen cluster to recipe, or change recipe for magic frame to use enchanting drop and increase it's cost for using a "less potent/appropriate magical material".
> 
> Other changes would need more discussion.

I think the main issue is that options 1 and 2 are pretty unreasonable, and this recipe is based on them.
Adding more bee products would make this recipe go against its original purpose.
Adding more tech or magic items doesn't actually make the recipe harder, since those items are useful on their own anyway.

Honestly, the whole bee system is a mess anyway. Maybe bees need a rework like crop. 
But for now, if there's no plan to nerf loot bags, I think this recipe is reasonable.

LazyFlesh:
I passed the proposal through meta dev, and currently it seems most people that responded are inclined to remove the soulful wax and the magical frame from the bag (remove the rng).

Imo, this should come with a buff to the durability of the mutagenic frame, so it's not dogshit to use enmasse.

Gives people the option to:
1. find oblivion in stronghold (3 per world, so at least 1 is basically guaranteed)
2. Do bees raw (stone to lv/mv before they have power for a WA or two) until they can get either alveary or other frame to make cycles faster
3. wait until they have some tech, and then do bees with mutagenic frames (Which if dura is increased won't be bad. They currently only offer 3 uses, since they don't work on their last durability point.)

I think this works well enough for tech tiering? Stone to steam is raw, lv has WA and some of the magic bees frames, mv has alveary and elec stims, hv has BM frames, industrial apiary, and eyes of ender to find oblivion frames, ev has mutagenic (pain to get), and iv to luv make mutagenic more reasonable. Obv WA is applicable to every tier.

This, if nothing else, lines up with how tech tiering allows skips if you're sufficiently above the entry threshold for the mod.

1328116700:
> I passed the proposal through meta dev, and currently it seems most people that responded are inclined to remove the soulful wax and the magical frame from the bag (remove the rng).
> 
> Imo, this should come with a buff to the durability of the mutagenic frame, so it's not dogshit to use enmasse.
> 
> Gives people the option to:
> 
> 1. find oblivion in stronghold (3 per world, so at least 1 is basically guaranteed)
> 2. Do bees raw (stone to lv/mv before they have power for a WA or two) until they can get either alveary or other frame to make cycles faster
> 3. wait until they have some tech, and then do bees with mutagenic frames (Which if dura is increased won't be bad. They currently only offer 3 uses, since they don't work on their last durability point.)
> 
> I think this works well enough for tech tiering? Stone to steam is raw, lv has WA and some of the magic bees frames, mv has alveary and elec stims, hv has BM frames, industrial apiary, and eyes of ender to find oblivion frames, ev has mutagenic (pain to get), and iv to luv make mutagenic more reasonable. Obv WA is applicable to every tier.
> 
> This, if nothing else, lines up with how tech tiering allows skips if you're sufficiently above the entry threshold for the mod.

I agree. The Mutagenic Frame should be buffed, as it currently only has 2 effective uses.
Even with the ability to repair or mass produce them, having to swap them out all the time is still annoying.
Also, the Necrotic Frame can indirectly be used to obtain a Magic Frame.

LazyFlesh:
> Also, the Necrotic Frame can indirectly be used to obtain a Magic Frame.

how so? does it decay into a magic frame when it dies?

1328116700:
> > Also, the Necrotic Frame can indirectly be used to obtain a Magic Frame.
> 
> how so? does it decay into a magic frame when it dies?

imageBecause of this quest

LazyFlesh:
closing in favour of pr with the proposed soultion: https://github.com/GTNewHorizons/GT5-Unofficial/pull/6926

PR #1790, opened 2026-06-04
Blackstone decorative recipes

EnderProyects:
- Added two recipes ( related to this issue https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/21433 )

- 64 cobble + 8 blocks worth of molten black granite = 64 blackstone
- 1 blackstone + 4 ingots worth of molten gold = 1 gilded blackstone ( 4 ingots worth because the block gives back gold nuggets when broken, and because of a mysterious Goose... yea, fortune stacks...)

### DIFF
--- a/src/main/java/com/dreammaster/scripts/ScriptEFR.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptEFR.java
@@ -39,6 +39,7 @@ import static gregtech.api.recipe.RecipeMaps.assemblerRecipes;
 import static gregtech.api.recipe.RecipeMaps.autoclaveRecipes;
 import static gregtech.api.recipe.RecipeMaps.cannerRecipes;
 import static gregtech.api.recipe.RecipeMaps.centrifugeRecipes;
+import static gregtech.api.recipe.RecipeMaps.chemicalBathRecipes;
 import static gregtech.api.recipe.RecipeMaps.chemicalReactorRecipes;
 import static gregtech.api.recipe.RecipeMaps.compressorRecipes;
 import static gregtech.api.recipe.RecipeMaps.cutterRecipes;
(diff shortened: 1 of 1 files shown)


LazyFlesh:
Do we have any capability to disable it's for tuning? It's a decorative block, so I'd rather it be cheap... Not that builders arent allowed to cheat in stuff for building purposes.

EnderProyects:
... I have no idea xd

LazyFlesh:
It does not stack. Tested in game.

Code overrides fortune to a max of 3, https://github.com/GTNewHorizons/Et-Futurum-Requiem/blob/master/src/main/java/ganymedes01/etfuturum/blocks/BlockGildedBlackstone.java

you get a max of 5 nuggest per block, and out of 18, I got 71. 

4 nuggets per block is sufficient for net negative returnss

src/main/java/com/dreammaster/scripts/ScriptEFR.java
```diff
@@ -876,6 +877,18 @@ public void loadRecipes() {
...
+        // BlackStone recipes
+
+        GTValues.RA.stdBuilder().itemInputs(new ItemStack(Blocks.cobblestone, 64, 0))
+                .fluidInputs(new FluidStack(FluidRegistry.getFluid("molten.graniteblack"), 1152)).circuit(4)
+                .itemOutputs(getModItem(EtFuturumRequiem.ID, "blackstone", 64, 0)).duration(30 * SECONDS)
+                .eut(TierEU.RECIPE_LV).addTo(assemblerRecipes);
+
+        GTValues.RA.stdBuilder().itemInputs(getModItem(EtFuturumRequiem.ID, "blackstone", 1, 0))
+                .itemOutputs(getModItem(EtFuturumRequiem.ID, "gilded_blackstone", 1, 0))
+                .fluidInputs(Materials.Gold.getMolten(576L)).duration(16 * SECONDS).eut(TierEU.RECIPE_LV)
```
LazyFlesh: INGOTS / 2 should be enough gold for the recipe. Fortune is capped on 3 for blackstone.

LazyFlesh (changes requested):

src/main/java/com/dreammaster/scripts/ScriptEFR.java
EnderProyects: done

LazyFlesh (commented):
Still no idea what the bug is, but if it exitsts, ill just assume you've found a good enough balance.

Processing time is a bit long, though, but ig it's lv. itll speed up.

lgtm otherwise.

LazyFlesh (approved):

PR #1792, opened 2026-06-05
Add Computation and Arithmetic Singularity Recipes + singularity deps

Auynonymous:
title, these recipes are giga expensive, to be used in the 2.9 stargate dial.

Only one will be needed for gate (10 if you make 10 😨)

Arithmetic Singularity
image

Circuit cost of 1 singularity
image
image

Recipe tested and works in Full Pack 👍 

Note: 1 ulv circuit is missing, this will be added to the dire table dial recipe :)

Also adds commented out deps for eternal and universal singularities, if someone ever needs to use them :)


Ruling-0 (approved):
I think the numbers are good.

PR #1800, opened 2026-06-08
reduce high tier material cost of foundry casing and chassis

chrombread:
t1 foundry was disproportionately expensive. this should bring it in line. i may buff the t1 pairing a little bit as well, as it feels like exofoundry is bait in UEV.

reduces optical components used by 25%.
reduces infinity cost by at least 33%. nuclear star is also reduced which helps
reduces uev supeconductor cost by 75%

this is only the base casing and t1 chassis, the other two parts are still expensive but I think this is a good start.


Auynonymous (approved):
looks goog

PR #1801, opened 2026-06-08
Fix Borosilicate Glass Ids in Coremod Recipes

Auynonymous:
title. adjusts foundry glass and nac glass to be uhv neutronium glass instead of cosmic neutronium (CN was removed) Also fixes waterline glasses.

Do not merge without gt5u pr :p
https://github.com/GTNewHorizons/GT5-Unofficial/pull/7025


Yoshy2002 (approved):
Balancewise i believe in auyna and maya, code is good
Dont merge without gt5 pr

PR #1808, opened 2026-06-09
Log warnings for null values instead of crashing in TCHelper

koolkrafter5:
Trying to modify a null research and it failing is not critical for the pack so it should not crash


Miklebe (approved):
Chirp chirp, a bird.

PR #1809, opened 2026-06-09
Add new recipes for updated dyson swarm module

flamingowrangler2869:
complementary to [this PR](https://github.com/GTNewHorizons/GT5-Unofficial/pull/7059)

This is to migrate away from weird dire-crafting table recipes with EV tier resources and towards more modern items/materials, appropriate for a UIV tier recipe

changes include:
+Corona-Resistant Drone Case
+Hyperfine TrNq+ Radioisomer Pellet
+Magnetically Shielded Femtocontroller
-dyson swarm module recipe change

java_J81yYrFTSF
qPp9LUdQFq
jjgqndwnfa
java_Ye6WBXCWeu
java_25wrjFFIxb


GDCloudstrike (changes requested):
This change would make dysons a lot more expensive to both start up and maintain, needs further discussion

EnderProyects:
Dysons are no longer needed in EoH recipes ( change I font like, but whatever), so this change is not as needed, also that if you are doing it for the power, with all the buffs to dire table this is not needed ( the fact that they can do 64 recipes at once)

flamingowrangler2869:
here's a comparison with the current recipe chain:

image

FourIsTheNumber:
> Dysons are no longer needed in EoH recipes ( change I font like, but whatever), so this change is not as needed, also that if you are doing it for the power, with all the buffs to dire table this is not needed ( the fact that they can do 64 recipes at once)

Dire table just sucks to use and is weird for these recipes. The goal here looks like it is to modernize and use more appropriate materials, which makes sense, because the current recipes are super weird.

flamingowrangler2869:
*i meant to write `radox casings: 32 -> 8`, not 4

EnderProyects:
The good thing about Dyson is that is "cheap" to maintain the modules, instead of doing a replace of the recipe just place this recipe on the space elevatir ( so it doesn't conflict with scan) and if you wanna do this recipe better, make so it outputs more than a stack

Auynonymous (commented):
The items themselves are ok, however the recipes do not account for how many dyson modules get destroyed, and how often they do. You need 10k modules to have a max eu dyson on horus. The inputs should be severely buffed here. Using EV materials was great imo. The only terrible parts of old dyson were the solar panels (thats whatever), Dire table nonsense, and End Powder.

Dream-Master:
@flamingowrangler2869 check why the action script is failing (get mod items need to be changed)

flamingowrangler2869:
am i able to switch it out for ItemList references yet? i'm new to cross-mod stuff sorry

UltraProdigy:
It can switch out now. The build will just fail until your GT5u is merged (& deps are updated here). You can make sure this builds with your PR build by taking advantage of your local maven.

flamingowrangler2869:
i got a bit lost.. hopefully i didnt screw anything up this time

GDCloudstrike (approved):

PR #1821, opened 2026-06-20
Remove Steel Dust Smeltery Melting

UltraProdigy:
This bypasses needing the BBF / EBF and thus should not be a thing. We already did this for the other smeltery accessible materials that require EBFing; this is an oversight.

### DIFF
--- a/src/main/java/com/dreammaster/scripts/ScriptTinkersConstruct.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptTinkersConstruct.java
@@ -1526,6 +1526,7 @@ public class ScriptTinkersConstruct implements IScriptLoader {
         TConstructHelper.removeMeltingRecipe(getModItem(TinkerConstruct.ID, "GravelOre", 1, 5));
         TConstructHelper.removeMeltingRecipe(getModItem(TinkerConstruct.ID, "materials", 1, 39));
         TConstructHelper.removeMeltingRecipe(GTOreDictUnificator.get(OrePrefixes.dust, Materials.Cobalt, 1L));
+        TConstructHelper.removeMeltingRecipe(GTOreDictUnificator.get(OrePrefixes.dust, Materials.Steel, 1L));
         TConstructHelper.removeMeltingRecipe(getModItem(TinkerConstruct.ID, "SearedBrick", 1, 2));
         TConstructHelper.removeMeltingRecipe(new ItemStack(GregTechAPI.sBlockOres1, 1, 382));
         TConstructHelper.removeMeltingRecipe(getModItem(TinkerConstruct.ID, "materials", 1, 38));


LazyFlesh:
While steel dust melting being a thing was weird, I thought it was nerfed by taking forever to be cast into ingots.

UltraProdigy:
Attaching a balance tag because it's changed behavior from status quo

Auynonymous:
perhaps we just adjust the cooling time by a lot more. steel dust is also slow to reform in bbf, but honestly this also feels wrong. getting to recycle steel from accidental parts using the steel dust is a good feature to have early.

UltraProdigy:
This only removes the dust to molten steel in the smeltery. People can still recycle tinkers parts and smelt GT parts back into the ingot.

It makes no sense for an EBF requiring material to be able to go from Dust -> Ingot in the smeltery. We already disallow this with Aluminium, Cobalt, Manyullyn, & Ardite

koolkrafter5:
> It makes no sense for an EBF requiring material to be able to go from Dust -> Ingot in the smeltery. We already disallow this with Aluminium, Cobalt, Manyullyn, & Ardite

You would be right if this material required the EBF. Since it can be made in the BBF, it seems plausible that the smeltery can also handle it. If you want to nerf it, the way to do it would be to tweak the melt time or cast time imo. Being able to use the smeltery to recycle steel dust from lathing and the like is part of what keeps LV bearable. You already need to make a BBF for the steam age quests iirc, so most players will still use it as their primary source of steel ingots.

Impos913:
It's not an oversight. It has been a feature for about 59 years and it was even nerfed to cool down slower. 

That would be a drastic change in early gameplay, especially for more experienced players. It would remove all recycling of looted steel tools and armor, chain armor that commonly drop from infernals. It would also destroy the passive steel from mob farms.

However, if you insist on this nerf, you must add new recipes to allow recycling of previously mentioned items and also speed up the steel dust to steel ingot recipe in BBF and reduce fuel cost.

UltraProdigy:
This is plainly not true. You can still use an arc furnace or fluid extractor to recycle steel tools and chainmail armor directly into ingots. This simply removes the bypass of being able to melt dust into ingots in the smeltery essentially for free. The BBF, EBF, and Arc Furnace have appropriate costs attached to them. 

As you've mentioned you can farm steel dust essentially for free with a mob farm or with CropsNH, having this extra bypass is ridiculous and does not fit in line with any of the other intended steel recipes.

Impos913:
Arc furnace is mid-LV and at that point you don't even need the 2 extra steel you might get from recycling some tools and armor you have been hoarding since stone age. 

Putting steel dust in BBF or EBF is never worth it since the time and cost is almost the same as making steel from iron.

Also I'd like to point out the effects of this on early gameplay again. For years I and many experienced players have used this recycling to obtain small amounts of steel pre-BBF to make the tool forge and make large tinkers tools. This allowed players to overcome the most tedious part of making the BBF, collecting resources by mining one block at a time.

UltraProdigy:
I'm not worried about causing an imposition to experienced players who have used this bypass to skip progression in the past. In fact that only bolsters my reasoning for getting rid of it. 

As I've mentioned already the recycling is still present, it just takes an appropriate amount of effort

Impos913:
The only things you will achieve with this change are making the early game more tedious and forcing one path of progression by removing options. 
Both of those things only hurt the pack.

Dream-Master:
> While steel dust melting being a thing was weird, I thought it was nerfed by taking forever to be cast into ingots.

Compared to bbf it's short .

Dream-Master (approved):

IX-git:
I have to say, I really dont like this change. The reason I play GT:NH is to optimize my playstyle and try weird solutions to get problems solved.

What im hearing is that steeleafranks, the steel crop, are too free and give far too much steel dust for an LV centrifuge. Those have been nerfed, and now need steeleaf blocks below them, so now they cant be easily ground out as they need an EBF. The main method other than steeleafranks is infernal mob farming or villager farming, both require a large investment in steam age to get to the chain armor, which then needs to be processed in a steam macerator, and then finally cast out in a tinkers smeltery at the cost of lava. 

This is an interesting and different choice a player can make to increase the difficulty of the investment but increase steel returns, which makes the game fun.

If you are set on removing this, please consider buffing the bbf recipe for steel dust to steel ingots, potentially making each ingot only take one minute to smelt from processed steel dust. This would still somewhat reward a player, but they would still have to craft out the BBF.

UltraProdigy:
The tinkers casting is still too cheap and doesn't make canonical sense however you have a good point that people doing extensive steam age tech like mob farms should generally be rewarded... though I will not say that it inherently needs it. 

Going through all that effort even without the free steel alongside gives plenty of rewards on it's own and as worth it with or without this change- iron, gold, diamonds, redstone, glowstone, ghast tears, spawn eggs, coal, pistons, ender pearls, anything craftable with Exp- which is a lot, etc all free- and with uncapped looting mind you.

I think allowing the BBF to smelt steel dust a bit faster is a fine compromise however. 

@Dream-Master what do you think about making steel dust in the BBF a little quicker than the other recipes. We could reduce it from 240s to 180s (aka 25% reduction based on fuel). I think as long as it stays at least 120s it won't compromise EBF balance.

akip1:
people aren't putting in the effort for steeleafranks/mob farms/villager farms so that they can continue to wait for 3-4 minutes per steel ingot, they're doing it explicitly to avoid that long wait, so continuing to force a long wait on players who found a solution is strange to me

its not like recycling steel dust is "too easy" of a solution to do, you're putting in _more_ effort than someone who just builds a BBF and leaves their pc on overnight, and all you get for that is the exact same speed. the steel dust you get from these other solutions is now as good as iron dust (and you can also get iron from these things), except you can only use it for steel

the idea that this is "skipping progression" is also just wrong imo, this is an alternative. you take a different path with different problems to get a different solution to the same problem. i'm not getting something i shouldn't have access to, i'm getting it in a different way. the idea that this doesn't make "canonical sense" i think is also wrong but it doesn't even matter. removing this alternative makes the gameplay strictly less interesting because you Just Build A BBF and theres nothing else to do but Make More of them if you want to have steel faster. even if you don't think it makes sense that a smeltery could melt steel, it removes an alternative solution to a problem. melting steel dust in a smeltery doesn't affect the BBF's place, because it still takes effort to make any of these farms and some players might not want to do that. some players also might not want to make a BBF because it takes 4 hours to make a stack of steel (assuming you use coal coke). including myself. using "canonical sense" as an excuse for this change doesn't make sense because like. you're not stopping me from melting a steel ingot in the smeltery. so theres clearly consideration where Gameplay takes precedence over if something Makes Sense, but why does that not apply to steel dust melting? steel dust melting is a _good_ thing for gameplay.

if you really think the method is too _good_, which i honestly think is fair, just nerf it in a way that doesn't outright remove it. you could make steel dust be half as efficient in a smeltery and i don't think people would be that mad about it. if you think that the amount of steel is too much, nerf the amount of steel, instead of removing an entire option for progression.

jackalope-bit:
To me, this change seems like it's trying to lock early gameplay into a single path. Its somewhat reasonable to make the change given that steel now always requires a carbon source to create, however given that getting the steel dust requires a whole other progression pathway (making a mob, setting up early game item handling) with its own challenges, it does seem like an unfair punishment to anyone wanting to take a different pathway

Looting is limited to 4 max now, so that argument is not valid. Additionally, XP autocrafting in steam and LV is a whole automation challenge on its own and /should/ be rewarded, because it requires significant investment (especially given the tier) of time, resources, and brainpower. 

There should be a reasonable steam age/early LV way to turn steel dust into steel ingots. 180s is not reasonable because it is already steel, and now you require fuel infrastructure on top of the mob farm. If steel dust could be smelted significantly faster than iron-> steel, and required less coal/coke, this would make sense. 

BBFs will never realistically compete with an EBF once available, this is about how the changes affect gameplay prior to LV. I don't think it necessarily needs to be returned to the smeltery, but if you are going to make the change to remove it, I think it's reasonable to provide an actual alternative, not just swing an axe at it because you feel like it.

I think the argument for making early game more painful has yet to be successfully made, and I'm disappointed to see you ignoring the feedback you're receiving here. It's a concerning trend lately, tbh. 

Hoping you reconsider.

OshiFantasy:
I get the idea of wanting to force BBF, but having no way for the player to recover steel if they turn a bit too much into dust when making obzinite dust or something seems a bit harsh. It's a sudden change to the player that only really happens at this specific progression point that turning their ingot into dust makes it completely unrecoverable

Ldmf:
The main problem with the change is simply that the costs for the rest weren't taken into account; as many have said before me, they just took a sledgehammer to the whole thing.

First, Steel is really only relevant in LV; after that, you need it less and less.
Second, they’re taking away the chance to find new approaches—and that’s really what it’s all about: finding a way to solve the problem (how you do it is up to you). But this way, they’re removing a path that was fun and required a different kind of thinking.

UltraProdigy:
Alright let's calm down a bit here. I will address your points individually.

But before I get to any of those comments I just want to make sure we're all on the same page. Steel is made in an EBF for 50s at MV from steel dust & 240-360s in a BBF (which, once again, I'm open to lowering within reason). The timeframe for smelting a steel dust is roughly a minute, 40 to melt & 20 to cast. This alone makes the BBF & EBF recipe virtually useless given the small free source of lava it runs on. Now consider not only that, but the fact that you can have hundreds of parallels on this by increasing the height of your smeltery and using a casting channel. You don't even need to create the bricks or the tables- you can just raid villages.

You see how this makes no sense? In essence, you're fighting for the ability to use an unintended steel creation method that costs no energy to run and can parallel into the hundreds. From the ripe age of steam you can have barrels of fully cooked up steel ingots for virtually no cost. That's not intended.

I completely understand the negative reactions, it was inevitable. If I were using a bypass this good (and I have before) I would be attached to it too. However, this is GTNH; it's a challenging pack. As a developer it's my role to look at things from a game design perspective and let go of my feelings as a player to make things make more sense.

As for the comments:
@akip1 
1) Yes building a mob farm or a villager farm will no longer provide you infinite steel ingots at the same ease. I'm not concerned about this. The effort you put into making one of those is already worthwhile without even considering the steel dust. As I listed up above here are some examples of mob farm drops: iron, gold, diamonds, redstone, glowstone, ghast tears, spawn eggs, coal, pistons, ender pearls, anything craftable with Exp- which is a lot, etc all free- and with uncapped looting mind you. All that and you still get infinite steel dust for free. There's just one extra step and it's leaving it to cook in the background while you complete other tasks, as is intended. There is no reason for building a farm to justify invalidating the intended early game steel bottleneck of the BBF.

2) The idea that this is "skipping progression" is 100% correct. “Alternative path” only holds if it still respects the steel gate. A route that turns farmed or looted dust into ingots without BBF/EBF cost is not just alternative, it bypasses the intended converter. No machines can melt down EBF requiring dusts besides blast furnaces is a normalcy in this pack.

3) Yes, I would be open to a minor reduction in BBF time to reward pre-created or dust recycled steel but allowing it in the smeltery doesn't make sense.

@jackalope-bit 
1) For reasons I've already stated this does not lock early gameplay to a single path. The current paths are still viable and worthwhile options for other reasons. In fact, it might even encourage more exploration because you can find lots of full steel ingots out in the world.

2) Looting is NOT limited to 4. I'm not sure where you're getting that information from. Technically speaking looting is capped at 32k as an enchantment that can be used on spikes. So once again, farming mobs is still viable. Also, we will just have to agree to disagree on calling a strategy that is well documented on YouTube and Discord a high brainpower operation.

3) This is a basic part of the challenge and the way that most people play the pack; without this. You'll find that once you reach LV there are plenty of ways you can take advantage of Fluid Extractors & Arc Furnaces to bypass the EBF anyways. If anything this merely pushes back the abuse to a more reasonable age.

4) This point is moot because both a BBF and an EBF cannot at all compete with the smeltery. Not even close. I am not "swinging an axe because I feel like it" and I resent the implication. It is a substantially, ridiculously superior method of steel production that does not fall in line with intended progression. I also protest your statement about this being a recent trend. 

@OshiFantasy 
1) We have an assortment of tools available so you can get the dust amount correct like NEI. Also it's still very much recoverable.

@Ldmf 
1) Steel is NOT only relevant in LV. Steel is (number wise) increasingly relevant with every tier. I don't know what you meant by this.

2) This is a relatively well known and documented bypass. Anyone looking through NEI for a few minutes can find it. I'm not saying we shouldn't reward that, but again, please consider the scale of the reward here; it's obscene.

Also, for the last three of you; GitHub is not meant to be an avenue of complaint. If you are going to provide feedback in the future please focus on making it constructive with actionable solutions and an open mind instead of outright dissent.

SafentisFox:
GTNH is a pack all about giving you different separate ways to progress that come together at certain points, but still leave you with a lot of flexibility, to the point where you could easily play multiple playthroughs without feeling that you’re just doing the same thing over and over again.

It is not a pack about forcing you towards one canonical path and looking at all other interactions as “oversights”. 

Are you going to remove steel nugget duping next? Perhaps you will remove the aluminium drop from Kobolds? After all, you’re “supposed” to use the EBF for it.

Are you going to remove tantalum nuggets duping next? It’s a rare material that you’re supposed to get through mining and processing ore, not by just throwing stuff at a cauldron! 

Meteors are kinda completely broken when you think about it, aren’t they? You can just completely skip using rockets with them! Rockets are supposed to be the way to progress.

I’m certain you can come up with appropriate points to rebuff me. Frankly though? It’s not relevant. All these small arguments and points people have made and you’ve responded to don’t matter at all. What matters is this, and I have not see it addressed neither in you original post nor in any following:

**How was the possibility of smelting steel dust materially harming the gameplay experience of GTNH’s playerbase?**

And I answer thus: **It was not**

Note that I’m not arguing whether it was intended or not. That’s irrelevant. It was in the game for a long while and the only valid reason to remove something is not if it’s “unintended” but rather when it’s harmful.

But there was no harm being done at all. You are not removing it as a way to improve the mod pack experience. You are removing something you personally perceive as “incorrect” using your own personal values, completely ignoring the fact that it did literally zero harm to the experience of GTNH’s players. Unlike its removal.

Could someone theoretically use this interaction to skip ever building the BBF? Yes. Much like many many other instances of similar skips within the mod pack. Could someone theoretically build up a system wherein a smeltery could create amazing amounts of steel, powered by an infernal farm, much faster than the EBF? Yes. By completely sacrificing progress towards further ages, by spending a lot of time on it, a player can successfully gain effectively infinite steel, a bit earlier than they would’ve otherwise been able to do the same thing.

How is this a bad thing? How does this negatively affect the game? It does not. There was a technically easier method to get a basic material, and at great quantities only if you sacrificed progress towards everything else, and you removed it for no reason. 

And I repeat, there’s many easy methods of gaining otherwise rare materials. Are you going to remove them too? You should. You must. If you stand by your actions that this was good for the game any and all other “easy methods” of obtaining materials must also be removed. 

Nugget duping of tantalum? Gone. Aluminium drops from kobolds, gone. Smelting platinum dust in a furnace into an ingot? Bah, IV EBF recipe. Plat line recipes that let you get platinum before an LCR? LCR only. 

Meteors that let you get materials of a planet before you’ve even been on that planet? Gone. Make them require the planet soil or more blood. 

And so on and so on and so on. There’s so many little details, these are just off the top of my head, little skips that are fun to discover and interact with. Little instances that make GTNH the great modpack that it is. You never feel trapped. You know that in the majority of cases there are other potential paths you can take.

That is why I fell in love with GTNH. Why unlike every single other modpack I tried, I stuck through with GTNH.

But it seems, it’s not the case for you. There is only one path and one path only that all players must take in your GTNH. And anything else is an “oversight” to be corrected, regardless of whether it negatively affects anyone or not. So because it’s something that you consider “unintended” it must go.

IX-git:
I think Fox put it very well, the choice is what makes GT:NH fun. And in the grand scheme of things, this steel recipe potentially only improves the feel of earlygame, which can only be good for all of the other content in the game.

I think that from a gameplay perspective, this change would make me and many others want to start a new GT:NH run a lot less, which is probably not good. I would love to see it get reverted potentially!

OshiFantasy:
If this choice is related to a nerf of witch farms being an easy source of steel and that's what needs to be addressed, **I propose instead removing the "always infernal" tag from witches**. There are blazes as an alternative farm source, but you are then either able to get a diamond dolly (and are in LV and can arc furnace chain anyways) or you are setting up automations in the ether, which is MUCH harder in steam. 

If you did this and allowed steel dust in the smeltery, steel dust wouldn't be functionally equivalent to or worse than iron dust (takes just as long in BBF, LONGER in ebf) and would allow this method of getting steel for the people that want to tackle the challenge of doing automated steam in the nether, which would hopefully be a much more acceptable level of difficulty for the dev team.

yuriks:
> steel dust wouldn't be functionally equivalent to or worse than iron dust (takes just as long in BBF, LONGER in ebf)

In my opinion, this is the actual problem with this change. Unless steel dust recipes are buffed, it is literally faster, *by a factor of about 10x*, to make steel ingots in the EBF starting from iron dust (via cast iron) rather than starting from steel dust. It doesn't make sense and feels like it just punishes trying to be economical with, e.g. lathes in LV.

UltraProdigy:
Thank you all for the feedback, it's much appreciated. We are taking it into account before the official 2.9 release. As it stands this specific change- the dust being taken out of the smeltery- is final and will not be reverted. That doesn't mean there won't be follow up changes to compensate for the effects of this however; those are in active discussion and any PR made will refer to this one.

Future non-dev discussion can take place in the discord or in a separate suggestion thread on the modpack repo. Feel free to ping me with any questions.

PR #1825, opened 2026-06-21
Added Item Hunger Amulet. A baubleized Healing Axe to let off-hand be…

BranTheBuildor:
A baubleized Healing Axe to let off-hand be free and as a stepping stone until the nano-booster. Texture may not be final as the animation could be better if it even should have one.


UltraProdigy:
This is neat but I don't understand where it fits in. Right now we have:

Healing Axe -> Cans / Talisman of Nourishment / Fruit of Gris -> Nanobooster

Which is (convenience wise) offhand -> hotbar/inv (+ small cost)-> amulet (+ power cost)

This would introduce an amulet with no upkeep cost- making it the best option hands down out of all of these and skipping the layers of progression in place.

BranTheBuildor:
> This is neat but I don't understand where it fits in. Right now we have:
> 
> Healing Axe -> Cans / Talisman of Nourishment / Fruit of Gris -> Nanobooster
> 
> Which is (convenience wise) offhand -> hotbar/inv (+ small cost)-> amulet (+ power cost)
> 
> This would introduce an amulet with no upkeep cost- making it the best option hands down out of all of these and skipping the layers of progression in place.

I was unaware of the items between the axe and the nanobooster. When I had the idea I just wanted to clear up the off hand or hotbar slot/inventory. I figured that the incentive of the nanobooster over this would be its health regeneration. The food replensihment could be adjusted to be worse than it aswell I suppose.

FourIsTheNumber:
The health regeneration of nanobooster is totally irrelevant since you are ZPM and never leave your base

PR #1826, opened 2026-06-21
Add space assembler recipe for neutronium compressor controller

GDCloudstrike:
This PR adds a space assembler recipe for the controller block of the neutronium compressor multi. Its almost an exact copy of the dire table recipe, except it has some added indalloy and the large plates are replaced with quad plates.
Also removed a commented-out recipe.

image


Auynonymous (approved):
much needed

Auynonymous:
@GDCloudstrike can you resolve this spotless fail?

PR #1831, opened 2026-06-28
Convert MAX circuit parts to use cosmic solder

Ruling-0:
Updates the MAX circuit (Planck) to use cosmic solder, at half as much as the previous living solder.


Auynonymous (approved):
this solder did not exist until after these were made. good catch

PR #1832, opened 2026-06-28
Reduce time on sunnarium alloys

Ruling-0:
BEC recipes now use these extensively, and the current times are excessively long. This reduces all of them by a factor of 8.


Auynonymous (approved):
Zyngen rework when :)

PR #1835, opened 2026-06-28
Add recipe for BEC LoS Hatch

Ruling-0:
Recipes for the hatches added in https://github.com/GTNewHorizons/GT5-Unofficial/pull/7022

Necessary for BEC multis.

image
image


Auynonymous (approved):
magic in my endgame????

PR #1836, opened 2026-06-28
BUFF RADOX

1328116700:
Because of the APIC Wafer changes, the Radox requirement for Pico Circuits has doubled, but neither Radox nor Bac Vats received any buffs. As a result, players now have to spam Bac Vats.

UMV CoAL and T4 QFT need roughly 4,000 Pico Circuits, equivalent to about 32,000 ingots of Radox. Meanwhile, 64 UMV Bac Vats only produce about 3,427 ingots of Radox per hour. This is a huge time wall.

This change would make the progression much smoother while also reducing the cost of UIV superconductors, making them more competitive.

QQ截图20260628101949

QQ图片20260628093926


boubou19:
I did the maths, and if you decide to recycle all the super heavy radox, the heavy radox and the diluted xenoxene, the amount of radox polymer ingots per hour you get with your setup goes to 3660 ingots/h.

the optical circuit board are better to be made with the kevlar (faster and costs less materials), so i'm assuming the radox cost in there is none.

That only leaves the radox polymer cost in the pico circuit. To build a QFT T4, one needs 236 continuous spacetime rippers, which are 4 UMV emitters each, themselves being 4 pico circuit each. that's 3776 pico circuits for them.

The UMV CoAL casings are 24 picos each (8 for the casing themselves, 16 hidden in the robot arms UMV) which makes it 1032 pico circuits for an UMV CoAL.

Now a Pico is 4 ingots of radox polymer for the encasement, and 4 ingots for the APIC wafers, so the total radox polymer cost is 38464 radox polymer ingots. This makes 10.6h of crafting with your 64 UMV bacterial vats.

With your proposed change, your setup would produce 5856 ingots/h, which makes the amount of radox polymer required by all these pico circuits to be crafted in 6.6h instead of 10.6h, we are on the same order of magnitude for the crafts.

The actual issue here is that we need the OSMDs everywhere, and that adds a good pressure on the xenoxene, much more than the radox polymer. If anything, we should buff the xenoxene production, by downtiering the radox polymere/xenoxene recipes in the QFT to tier 3. The proposed changes does not solve anything.

1328116700:
> I did the maths, and if you decide to recycle all the super heavy radox, the heavy radox and the diluted xenoxene, the amount of radox polymer ingots per hour you get with your setup goes to 3660 ingots/h.
> 
> the optical circuit board are better to be made with the kevlar (faster and costs less materials), so i'm assuming the radox cost in there is none.
> 
> That only leaves the radox polymer cost in the pico circuit. To build a QFT T4, one needs 236 continuous spacetime rippers, which are 4 UMV emitters each, themselves being 4 pico circuit each. that's 3776 pico circuits for them.
> 
> The UMV CoAL casings are 24 picos each (8 for the casing themselves, 16 hidden in the robot arms UMV) which makes it 1032 pico circuits for an UMV CoAL.
> 
> Now a Pico is 4 ingots of radox polymer for the encasement, and 4 ingots for the APIC wafers, so the total radox polymer cost is 38464 radox polymer ingots. This makes 10.6h of crafting with your 64 UMV bacterial vats.
> 
> With your proposed change, your setup would produce 5856 ingots/h, which makes the amount of radox polymer required by all these pico circuits to be crafted in 6.6h instead of 10.6h, we are on the same order of magnitude for the crafts.
> 
> The actual issue here is that we need the OSMDs everywhere, and that adds a good pressure on the xenoxene, much more than the radox polymer. If anything, we should buff the xenoxene production, by downtiering the radox polymere/xenoxene recipes in the QFT to tier 3. The proposed changes does not solve anything.


> I did the maths, and if you decide to recycle all the super heavy radox, the heavy radox and the diluted xenoxene, the amount of radox polymer ingots per hour you get with your setup goes to 3660 ingots/h.
> 
> the optical circuit board are better to be made with the kevlar (faster and costs less materials), so i'm assuming the radox cost in there is none.
> 
> That only leaves the radox polymer cost in the pico circuit. To build a QFT T4, one needs 236 continuous spacetime rippers, which are 4 UMV emitters each, themselves being 4 pico circuit each. that's 3776 pico circuits for them.
> 
> The UMV CoAL casings are 24 picos each (8 for the casing themselves, 16 hidden in the robot arms UMV) which makes it 1032 pico circuits for an UMV CoAL.
> 
> Now a Pico is 4 ingots of radox polymer for the encasement, and 4 ingots for the APIC wafers, so the total radox polymer cost is 38464 radox polymer ingots. This makes 10.6h of crafting with your 64 UMV bacterial vats.
> 
> With your proposed change, your setup would produce 5856 ingots/h, which makes the amount of radox polymer required by all these pico circuits to be crafted in 6.6h instead of 10.6h, we are on the same order of magnitude for the crafts.
> 
> The actual issue here is that we need the OSMDs everywhere, and that adds a good pressure on the xenoxene, much more than the radox polymer. If anything, we should buff the xenoxene production, by downtiering the radox polymere/xenoxene recipes in the QFT to tier 3. The proposed changes does not solve anything.

Yeah, Xenoxene is the real issue. I can't discuss it with the other devs, so I'm not sure whether the numbers are balanced.
But Xenoxene is just really bad. Might as well be as generous as possible before Artificial Organisms.

PR #1838, opened 2026-06-30
Fix Unstable Ingots not exploding on shift click craft

FourIsTheNumber:
For some reason we are removing the unstable ingot recipe and then replacing it with an identical recipe (except worse, because the original recipe allows crafting in all 3 columns, while the coremod one allows only the leftmost column).

More importantly, this is the cause of the beloved bug where shift clicking the recipe output makes them not explode. Extra Utilities has its own custom recipe class and this replaces it with a generic coremod recipe class. The fix is just to use the ExU recipe as is, since we don't even modify it...

To all players - I am sorry.


Ethryan (approved):
Crafting Table go Boom.

Dream-Master:
To understand it right. Now the recipe work liken I intended?

FourIsTheNumber:
> To understand it right. Now the recipe work liken I intended?

Yes. The ingots now explode on a timer even if you shift click the output. Just a very simple exploit fixed. The recipe is also the same.

Jasonligit:
nbc

PR #1839, opened 2026-06-30
Add armor molds to solidifier hatch phantom slots

### LINKED ISSUE #25334: Add armor molds to phantom solidifier hatch slots?

dosipov

### Your Pack Version

2.9 beta 1

### Your Proposal

Image
As you can see, player (me) forced to create a subnet to add boots pattern in this specific case.
I believe game experience will be enhanced if that feature is present

### Your Goal

To make solidifier hatches more flexible

### Your Vision

Just QoL. If other molds are present, why not add those?

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this change please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [x] I understand this change request may not attract enough attention and thus not be implemented.
- [x] I understand this change request may be rejected due to other community members think it's inappropriate.
- [x] I believe this feature would make the pack better.

loenaaaa:
This PR adds the ability to use the 4 armor molds added in coremod in the GT Solidifier Hatch phantom slots. Note that it relies on the PR https://github.com/GTNewHorizons/GT5-Unofficial/pull/7312, so please do not merge this PR before that PR is merged. Additionally, it was tested using maven local, and not only can the molds be selected, but they can also be used in the Mass Solidifier (I tested using a dummy recipe):

Screenshot 2026-07-01 at 1 33 58 AM

This doesn't exactly affect balance since those molds don't appear to have any recipes in the fluid solidifier?

Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/25334.


UltraProdigy (commented):
This is fine, but instead of putting it in MainRegistry maybe instead give it its own helper in gthandler or something and call it in GT_CustomLoader near the recipe stuff

UltraProdigy (approved):
Looks great!

PR #1843, opened 2026-07-03
CropsNH Lootbag Enchanting Recipes

Dynamiczbee:
## Summary
Simple (in theory) PR just to register and create recipes for the two new CropsNH lootbags and deregister the old one.

## Checklist
- [x] Merge alongside https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/25643

### DIFF
--- a/src/main/java/com/dreammaster/scripts/ScriptEnhancedLootBags.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptEnhancedLootBags.java
@@ -19,6 +19,8 @@ import static gregtech.api.enums.Mods.Railcraft;
 import static gregtech.api.enums.Mods.Thaumcraft;
 import static gregtech.api.enums.Mods.ThaumicTinkerer;
 import static gregtech.api.enums.Mods.Witchery;
+import static gregtech.api.enums.Mods.CropsNH;
+
 
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


src/main/java/com/dreammaster/scripts/ScriptEnhancedLootBags.java
```diff
@@ -105,6 +108,9 @@ public void loadRecipes() {
         metaExtraItemMap.put(56, getModItem(Witchery.ID, "leonardsurn", 1, 3));
         metaExtraItemMap.put(57, getModItem(HardcoreEnderExpansion.ID, "enderman_head", 1, 0));
         metaExtraItemMap.put(58, getModItem(HardcoreEnderExpansion.ID, "rune", 1, 5));
+        metaExtraItemMap.put(59, getModItem(CropsNH.ID, "plantLens", 1, 5));
+        metaExtraItemMap.put(60, getModItem(CropsNH.ID, "spade", 1, 5));
```
C0bra5: The meta value (the last parameter) should be 0, not 5; these are currently uncraftable because of this.

As for the higher-tier loot bag, I would rather we use a higher-tier item than a spade since it's basically just 9 steel and a stick. Either up it to a reinforced spade or use something a bit harder to get, like plant cure, which requires a brewing mechanism and an assembler to make.

C0bra5 (dismissed):

src/main/java/com/dreammaster/scripts/ScriptEnhancedLootBags.java
Dynamiczbee: Inserted plantLens to ID 51 to match new Farmhand bag replacing old Crops bag. Also swapped spade for plantCure.

Dynamiczbee:
Should be good to merge tomorrow in parallel with the other PR @C0bra5

C0bra5 (approved):
LGTM

PR #1844, opened 2026-07-05
Readd Sunnarium Autoclave Recipes

EnderProyects:
## Summary
Add back the advanced solar panels items recipes in autoclave that where removed in this pr https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/1475 , addresing the part that the recipes where in batch of 64, dividing the cost so the recipe is now in a 1 to 1 item instead of 64 to 64 ( also dividing the time)

## Checklist

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,20 +1,20 @@
 // Add your dependencies here
 
 dependencies {
-    api("com.github.GTNewHorizons:GT5-Unofficial:5.09.54.06:dev")
+    api("com.github.GTNewHorizons:GT5-Unofficial:5.09.54.15:dev")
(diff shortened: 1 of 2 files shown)


Dream-Master:
@EnderProyects can you provide us a picture of the recipe?

boubou19:
The proposed recipes in this PR:
image
This is a reduction by 16, limited by the output. Previous time: 2min, new recipe time: 8s. This is not the proper time, it should be 150 ticks (`7*SECONDS+10*TICKS`)
image
This is a reduction by 64, like announced in the PR description. However the time has been divided by 16 and rounded, like the previous recipe. It should be 38 ticks if we apply the division by 64.

boubou19 (approved):

PR #1855, opened 2026-07-07
Rework Angel Ring Recipe

FourIsTheNumber:
## Summary
Result of proposal/vote: https://discord.com/channels/181078474394566657/1438902363774779452/1523996047636959284

The goal is to turn Angel Ring from a meme item into a capstone flight option for magic integration. Instead of random LV/MV/HV items, the recipe now takes magical items from many primary magic mods: Ichorium, Elven Elementium, Aether (TC, Botania, Blood Magic, Witchery by proxy). It also takes Mobius ingots, integrating ExU by requiring the end ritual (soon to be improved by UiE).

This moves the ring from ZPM to IV. The rationale for this change is that magic options should be available earlier than their tech equivalents, and there are already several means of getting lasting creative flight through magic at IV. Developers can read more by following the context link from the vote linked above.

image
image
image
image
image
image

Additionally:
-Rewrote the research pages. Sadly, I erased the old RU lang because it is no longer correct and I don't speak Russian, so this will have to be retranslated based on my new page
-Removed the research that does nothing but lead to the angel ring page
-Moved the research from Artifice to New Horizons (as per dev doc, and in keeping with other ExU researches)
-Made the cosmetic variants wildcard meta arcane crafting recipes instead of one-way infusions

## Checklist


Ethryan:
Hmm..🤔

Any thoughts about adding the Time or Space aspects as a requirements for crafting?

For some reason I’ve always considered the Angel Rings ability to in part be related to magically tweaking your position in the space-time in real time.

It might give those rarer aspects a bit more use.

FourIsTheNumber:
> Hmm..🤔
> 
> Any thoughts about adding the Time or Space aspects as a requirements for crafting?
> 
> For some reason I’ve always considered the Angel Rings ability to in part be related to magically tweaking your position in the space-time in real time.
> 
> It might give those rarer aspects a bit more use.

I'd have to see what the sources are. Personally I hate most of the custom aspects (I went on a rant yesterday in metadev lol) so I'm not naturally inclined to use them. If they have reasonable sources then I could add them.

FourIsTheNumber:
Tempus:
image

I don't know of any source for Caelus.

LazyFlesh:
Caelus is just enderperls afaik

FourIsTheNumber:
> Caelus is just enderperls afaik

Ah yea. That's doable then. I don't think Tempus makes as much sense personally so I will just add Caelus for now

koolkrafter5 (approved):
I like these new recipes. They give a reason to actually make this item, but it's not strictly better than other options on the tier (ichor chest, primordial armor).

Caelum is a good addition. Maybe we should try to add some more sources other than ender pearls in the future. CropsNH Space Flowers would be a good one to give it to.

PR #1856, opened 2026-07-08
Add Assembler Recipes For Some Compressors

fehling135:
## Summary
Involving EV compressor, UIV compressor, multi compressor cntroller and (advanced) implosion compressor controller. These are used a lot in black-hole which gets into EoH block recipe in 2.9. So I think it necessary to add these to assembler.
(wrong piston tier in electric compressor pic, fixed in code)
2adf15cabc2803c2c092640d245daaa0
c210009e5aa06601c044b5df36ef4ffe
60f60f27f09cfb005cb1d3bb0f974af4
8cb4e5214fc92f90c6bc122d697690e9
3ff312bb19a31bd447a3a83c3391f23a

## Checklist

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
@@ -2514,7 +2514,7 @@ public class AssemblerRecipes implements Runnable {
                         GTOreDictUnificator.get(OrePrefixes.circuit, Materials.EV, 2),
                         ItemList.Hull_EV.get(1),
                         GTOreDictUnificator.get(OrePrefixes.cableGt01, Materials.Aluminium, 4))
-                .itemOutputs(ItemList.CompressorEV.get(1)).duration(5 * SECONDS)
+                .itemOutputs(ItemList.Machine_EV_Compressor.get(1)).duration(5 * SECONDS)
                 .eut(TierEU.RECIPE_EV).addTo(assemblerRecipes);
 
         // UIV Compressor
@@ -2530,7 +2530,7 @@ public class AssemblerRecipes implements Runnable {
         // Industrial Compressor
         GTValues.RA.stdBuilder()
                 .itemInputs(
-                        ItemList.CompressorEV.get(1),
+                        ItemList.Machine_EV_Compressor.get(1),
                         GTOreDictUnificator.get(OrePrefixes.circuit, Materials.IV, 2),
                         ItemList.Electric_Piston_IV.get(2),
                         GGMaterial.incoloy903.get(OrePrefixes.plate, 4))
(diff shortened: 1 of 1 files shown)


src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
```diff
@@ -2507,6 +2517,49 @@ public void run() {
...
                 .addTo(assemblerRecipes);
 
+        // UIV Compressor
+        GTValues.RA.stdBuilder()
+                .itemInputs(
+                        ItemList.Electric_Piston_UIV.get(2),
+                        GTOreDictUnificator.get(OrePrefixes.circuit, Materials.UIV, 2),
+                        ItemList.Hull_UIV.get(1),
+                        GTOreDictUnificator.get(OrePrefixes.cableGt01, Materials.NetherStar, 4))
+                .itemOutputs(ItemList.CompressorUIV.get(1)).duration(10 * TICKS)
```
UltraProdigy: Why so quick?

Nana-Sakura: isn't it originally workbench recipe? not bad ig

Ruling-0 (dismissed):
Thanks for the PR! Another thing off my TODO list 🎉. Just run spotless so the checks pass.

fehling135:
wait i mistake piston in electric compressor to IV
lemme fix it

src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
Ruling-0: It should be fine at 5s, matching the others.

fehling135: But if we make it 5s it has no competitiveness with workbench recipe. Its processing voltage is UIV.

UltraProdigy: Honestly I'm shocked they're bench recipes in the first place being so high tier.

Ruling-0: Yeah that's a fair point, this is basically the highest level recipe that fits this case. Ran the numbers and it does register (barely) within a gate craft. How about 1s?

Nana-Sakura: 1 sec is reasonable imo

fehling135: yeh i also think it acceptable

Ruling-0 (approved):

PR #1857, opened 2026-07-08
reduce the carbon dioxide output from steel production in the blast furnace

ABKQPO:
## Summary
producing 1000 of carbon dioxide allows players to generate more oxygen through steel production, creating an infinite oxygen loop. reducing it to 250 fixes this issue

## Checklist
- [ ] This PR is in compliance with the [GTNH AI Policy](https://github.com/GTNewHorizons/GTNH-Dev-Doc/blob/master/AI_POLICY.md)


luoyangyuli2001:
I love 250, good number.

Side note: had the old EBF muffler-recovery mechanic (removed in #3677) still been around, this loop would only break even at an IV muffler and only become positive from LuV up. But this recipe was added after #3677.

koolkrafter5 (approved):

PR #1861, opened 2026-07-11
fix hungarian notation

MLGfruitshoot:
goes with https://github.com/GTNewHorizons/GT5-Unofficial/pull/7365

just renames the vRadiationLevel to radiationLevel


C0bra5 (approved):
LGTM

PR #1870, opened 2026-07-15
Add tartarite to alpha Centauri Bb stone dust products

fehling135:
## Summary
Currently NAC controller needs celestial tungsten plasma, which can only be made in fusion and requires tartarite. But for rocketless players they can only get tartarite after DTPF, with space miner mkIII meteor. 
This PR fixed the problem by adding tartarite to alpha Centauri Bb stone dust products. So players can get the stone through meteor and reach tartarite befor DTPF.
(At first I changed NAC controller recipe, but reverted later. See downside discussion.)

## Checklist
- [ ] This PR is in compliance with the [GTNH AI Policy](https://github.com/GTNewHorizons/GTNH-Dev-Doc/blob/master/AI_POLICY.md)

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/CentrifugeRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/CentrifugeRecipes.java
@@ -381,7 +381,7 @@ public class CentrifugeRecipes implements Runnable {
                         GTOreDictUnificator.get(OrePrefixes.dust, Materials.Neutronium, 9L),
                         GTOreDictUnificator.get(OrePrefixes.dust, Materials.BlackPlutonium, 4L),
                         getModItem(Avaritia.ID, "Resource", 36, 2),
-                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Tartarite, 4L))
+                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Tartarite, 1L))
                 .outputChances(5000, 2000, 500, 2500, 150, 80).fluidOutputs(Materials.Mercury.getFluid(7200L))
                 .duration(6 * MINUTES + 28 * SECONDS + 16 * TICKS).eut(TierEU.RECIPE_LuV).addTo(centrifugeRecipes);
 
--- a/src/main/java/com/dreammaster/gthandler/recipes/CentrifugeRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/CentrifugeRecipes.java
@@ -381,7 +381,7 @@ public class CentrifugeRecipes implements Runnable {
                         GTOreDictUnificator.get(OrePrefixes.dust, Materials.Neutronium, 9L),
                         GTOreDictUnificator.get(OrePrefixes.dust, Materials.BlackPlutonium, 4L),
                         getModItem(Avaritia.ID, "Resource", 36, 2),
-                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.InfinityCatalyst, 4L))
+                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Tartarite, 4L))
                 .outputChances(5000, 2000, 500, 2500, 150, 80).fluidOutputs(Materials.Mercury.getFluid(7200L))
                 .duration(6 * MINUTES + 28 * SECONDS + 16 * TICKS).eut(TierEU.RECIPE_LuV).addTo(centrifugeRecipes);
 
--- a/src/main/java/com/dreammaster/scripts/ScriptNanochipRecipes.java
+++ b/src/main/java/com/dreammaster/scripts/ScriptNanochipRecipes.java
@@ -216,7 +216,7 @@ public class ScriptNanochipRecipes implements IScriptLoader {
                         ItemList.Sensor_UEV.get(4), ItemList.Emitter_UEV.get(4),
                         getModItem(OpenComputers.ID, "screen3", 1, 0), getModItem(OpenComputers.ID, "keyboard", 1, 0) },
                 new FluidStack[] { MaterialMisc.MUTATED_LIVING_SOLDER.getFluidStack(256 * INGOTS),
-                        new FluidStack(MaterialsElements.STANDALONE.CELESTIAL_TUNGSTEN.getPlasma(), 64 * INGOTS),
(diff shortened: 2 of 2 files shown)


LazyFlesh:
While I can agree with the change, both admins and devs have stated that GTNH is balanced around the base pack, not challenge runs. 

Instead, I think taking a look at rebalancing the endgame meteors a bit to bring them more in line to where they used to sit before all the content was inserted between the normal source and them would be a better call. Tengam, for example, you need a lot more before you get the SE MK3 miner.

I.e. you could make the t8 ores meteor (Your resources, hand them over, I think? Or Cheaty Late Game Ores meteor, if i remember the names right.) give a little bit of tartarite, as it spawns as an ore on a t8 planet.

fehling135:
> While I can agree with the change, both admins and devs have stated that GTNH is balanced around the base pack, not challenge runs.
> 
> Instead, I think taking a look at rebalancing the endgame meteors a bit to bring them more in line to where they used to sit before all the content was inserted between the normal source and them would be a better call. Tengam, for example, you need a lot more before you get the SE MK3 miner.

Yeh I have also thought of rebalancing meteor. But I cant find a proper item that has a similar cost to miner mk3 and available before DTPF. So I have to change the recipe itself. 
For tengam I think no large changes needed, it has cluster in 2.9 so u can get about 2 times of tengam from a single meteor, and the tengam cost before miner mk3 is about 3 times than before (if my memory doesn't go wrong). I think it acceptable.

fehling135:
> While I can agree with the change, both admins and devs have stated that GTNH is balanced around the base pack, not challenge runs.
> 
> Instead, I think taking a look at rebalancing the endgame meteors a bit to bring them more in line to where they used to sit before all the content was inserted between the normal source and them would be a better call. Tengam, for example, you need a lot more before you get the SE MK3 miner.
> 
> I.e. you could make the t8 ores meteor (Your resources, hand them over, I think? Or Cheaty Late Game Ores meteor, if i remember the names right.) give a little bit of tartarite, as it spawns as an ore on a t8 planet.

I have just gone to check the recipes, no tartarite from T8 stone dusts.
image

LazyFlesh:
Not the stone dusts, but the mateor containing the t8 ores, like inf cat.

Adding tartarite to one of the stone dusts instead of another, much less useful dust is another viable option, however, with much smaller balance implications.

Additionally, if the ore blend of the mo3 miner meteor is tweaked a bit, it wouldn't be unreasonable to make it use a mk2 miner instead. The cost is a lot less, but if there's nothing novel for a normal run (tartarite is vmable on a planet, for example) it should fall within the guidelines by which previous meteors have been added.

Auynonymous (dismissed):
Discussed with Maya, if celestial is changed to molten, Astral titanium should also become molten, for consistency.

Auynonymous:
I personally think it should stay plasma, and a better way of obtaining tartarite is added for gog.

LazyFlesh:
> I personally think it should stay plasma, and a better way of obtaining tartarite is added for gog.

This is actually just for no space/rocket.

We allow you to go to dims in gog. Once you have the SE, you can to tp the planet that has tartarite and vm it there.

Auynonymous:
> > I personally think it should stay plasma, and a better way of obtaining tartarite is added for gog.
> 
> This is actually just for no space/rocket.
> 
> We allow you to go to dims in gog. Once you have the SE, you can to tp the planet that has tartarite and vm it there.

fair enough. I think a meteor for it is the best solution. closest to space without space imo.

LazyFlesh:
> Additionally, if the ore blend of the mo3 miner meteor is tweaked a bit, it wouldn't be unreasonable to make it use a mk2 miner instead. The cost is a lot less, but if there's nothing novel for a normal run (tartarite is vmable on a planet, for example) it should fall within the guidelines by which previous meteors have been added.

The non-ion jet is also an option.

fehling135:
image

What if we change inf cat in alpha centauri Bb stone dust to tartarite? tartarite appears in samarium vein on this planet so its stone dust should contain tartarite, and the stone can be achieved through meteor.

LazyFlesh:
That's fine, imo. Like I'd said, it'd be minimal balance concerns compared to the other changes. I doubt anyone centrifuges the stone dust for the inf cat, either.

fehling135:
Thanks! Gotta make a change

fehling135:
Adjusted so about 1.5 meteor just enough for NAC controller.

src/main/java/com/dreammaster/gthandler/recipes/CentrifugeRecipes.java
```diff
@@ -381,7 +381,7 @@ public void run() {
                         GTOreDictUnificator.get(OrePrefixes.dust, Materials.Neutronium, 9L),
                         GTOreDictUnificator.get(OrePrefixes.dust, Materials.BlackPlutonium, 4L),
                         getModItem(Avaritia.ID, "Resource", 36, 2),
-                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Tartarite, 4L))
+                        GTOreDictUnificator.get(OrePrefixes.dust, Materials.Tartarite, 1L))
```
LazyFlesh, line 384: 4 was fine, imo.

 Though, It makes little difference either way, I guess.

fehling135: As spinmatron exists, centrifuging stone dust is rather easy. But yeah this doesn't affect a lot.

Auynonymous (approved):

Auynonymous:
@fehling135 update pr description so dream knows what this actually changes and why 👍

LazyFlesh:
I don't believe balance is affected anymore by this change.

Auynonymous:
> I don't believe balance is affected anymore by this change.

still technically changes balance, tag should stay

Auynonymous:
(he's gonna merge it regardless :trollface:)

Dream-Master:
I am not ingame but this change smell like a buff. Why it is  necessary ?

Auynonymous:
> I am not ingame but this change smell like a buff. Why it is necessary ?

celestial tungsten cannot be obtained without dtpf in no rocket runs. making nac not craftable until then. this pr adds tartarite (a material that can be void mined on this planet btw) to the stone dust, so it can be obtained using BM meteors before dtpf, so no rocket players dont have to wait until after dtpf for the nac.

boubou19:
regular playthroughs will just voidmine it. Fine to merge.

PR #1872, opened 2026-07-16
Enable subtypes for vanilla paper

glowredman:
## Summary
This sets the `hasSubtypes` flag of paper to true, allowing item stacks with meta values other than 0. This is used for the legacy Star Gate trophies, which previously did not work correctly.

## Checklist
- [x] This PR requires another PR in order to merge


Caedis:
Wouldnt it be better to change them to a dedicated item like the noob cert?

koolkrafter5 (approved):

koolkrafter5:
Unique items would be fine too if you want to change them, but this is fine otherwise

Caedis:
yeah i am fine with either, would just prefer dedicated items over adding meta support to vanilla items

UltraProdigy:
We can potentially switch this later I need the quest changes to merge now

PR #1879, opened 2026-07-26
Keep Warding Focus in Arcane Crafting Grid

koolkrafter5:
This removes a lot of the friction caused when bulk crafting Warded Glass; see the difference when tested in the latest daily version:

https://github.com/user-attachments/assets/04ba611a-e254-40ac-a84f-e14f74ebf689

https://github.com/user-attachments/assets/4c994541-570e-44ab-b2b3-ea888f5bc229

Don't mind the lag spike, that's normal for big batches of arcane crafts due to checking the wand's vis for every individual craft.

Also simplifies how the container item was set (removing that ugly ASM transformer and associated config) and sets the stack size to 0 to show the focus isn't consumed in the recipe when you bookmark it:
image

The BOP block stuff still works too after moving its mixin registration to account for this mixin.


src/main/java/com/dreammaster/coremod/DreamCoreMod.java
UltraProdigy: I believe this can be yeeted

src/main/java/com/dreammaster/mixin/mixins/late/MixinFocusWarding.java
```diff
@@ -0,0 +1,17 @@
...
+
+import net.minecraft.item.ItemStack;
+
+import org.spongepowered.asm.mixin.Mixin;
+
+import thaumcraft.api.wands.ItemFocusBasic;
+import thaumcraft.common.items.wands.foci.ItemFocusWarding;
+
+@Mixin(value = ItemFocusWarding.class)
+public class MixinFocusWarding extends ItemFocusBasic {
```
UltraProdigy, line 11: Should add an override
```
public ItemStack getContainerItem(ItemStack stack) {
    return stack.copy();
}
```
so that it copies over focus NBT

src/main/java/com/dreammaster/coremod/DreamCoreMod.java
koolkrafter5: https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/1879/commits/72d2338632eb6d25820607fa963a69ca1e02180a

src/main/java/com/dreammaster/mixin/mixins/late/MixinFocusWarding.java
koolkrafter5: https://github.com/user-attachments/assets/a7b8dea9-4b9d-4b2b-ab07-7ab130b1c3fe

https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/1879/commits/2f856588f102e9ca65f7f3a8c45a44ddca84bb8e

UltraProdigy (approved):
Looks good

PR #1901, opened 2026-08-10
Enhance stellar catalyst tesseract recipe

fehling135:
## Summary
At now tesseract's stellar recipe has 4x EU cost, same time but only 2x output compared with exotic one. This PR cut down time cost by 2x. Also resplendent recipe time is doubled. So that all the three recipes should have same speed with convergence.

## Checklist

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/recipes/DTPFRecipes.java
+++ b/src/main/java/com/dreammaster/gthandler/recipes/DTPFRecipes.java
@@ -432,7 +432,7 @@ public class DTPFRecipes implements Runnable {
                 // V3
                 GTValues.RA.stdBuilder()
                         .itemInputs(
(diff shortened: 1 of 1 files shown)


Dream-Master:
Why always recipes get buffed. Gtnh becomes to easy 😭

Caedis:
We have to have buffs to offset the nerfs

fehling135:
NERFS ARE COMING

Nana-Sakura (approved):
balance should be good, in the past we have resplendent recipe faster than exotic, and exotic faster than stellar, which is not good enough
now the resplendent is nerfed, and this change is ignorable for early stage dtpf users, but more in line with the fuel level for later tiers especially with converge

GDCloudstrike (changes requested):
The recipe time change is probably fine, but the stellar recipe being less efficient on transcendent metal was an intentional choice

GDCloudstrike (approved):

PR #1920, opened 2026-08-22
Dyson Module Rebalance and Minor Dyson Retier

Auynonymous:
## Summary
Remakes the Dyson Swarm Module recipes to be way less expensive. The recipes have become more similar to their 2.8 versions, but includes the new dyson parts and some uiv items (but in way less quantity than before)
Adds some SPASS recipes for "Creative" OpenComputer items that are used in dyson modules (drone case 3, microcontroller 3, and apu)

Switches out the UMV (pico) circuits in dyson blocks to be UIV (optical), doubles the circuit cost, and adds the radox cost from the umv circuits back into the recipes.
This change shifts dyson's tier placement from the tailend of UIV, to the start. Unlocked right after UIV components. All other costs are the exact same

This change is mostly because pico circuits will be removed and replaced with exotic circuits in 2.10, making dyson locked behind godforge, which is wayy too late for dyson to be effective. Plus it has to fight with antimatter and naqfuel for that placement. This change lets dyson be used for TPM power costs early (if you wanna facetank the heafty dyson cost) or for DTPFs. Naqfuel is quite powerful during UIV, sometimes making dyson not at all worth the effort (since you can just expand naqfuel. This should hopefully let dyson be more of a viable option for those who seek it)

A big cost of dyson modules will be the shifted superfluid, these numbers can be adjusted if needed, however the t2 recipe is unlocked with hypogen coils, which is a good upgrade from the starter recipe.

also has a gt5u pr to change the drone case texture to be more accurate: https://github.com/GTNewHorizons/GT5-Unofficial/pull/7873

<details>
<summary>Current 2.9 Cost vs New Cost</summary>

10k modules
image
(rounds up the cost cause of 64 items per craft)
image

</details>

<details>
<summary>Creative Item Recipes</summary>
image
image
image
</details>

<details>
<summary>Dyson Module Recipes</summary>
image
image
image
image
image
</details>

<details>
<summary>Dyson Block Recipes</summary>
image
image
image
image
</details>

## Checklist
- [x] I have tested this PR in Fullpack (Daily 702)


EnderProyects:
Is a small change, but I would have placed the Dyson solars in space assembler, and the battery that uses "the core" in a precise assembler, apart of that all look Ms good and is a reasonable price

Auynonymous:
> Is a small change, but I would have placed the Dyson solars in space assembler, and the battery that uses "the core" in a precise assembler, apart of that all look Ms good and is a reasonable price

fuel pellet is in assembler because plutonium pellet is. it's a thematic upgrade so I made it the same. as for dyson modules it takes aal infrastructure away, so I think keeping it here is fine. maybe if space assembler ever gets a rework

boubou19 (approved):
Seems decent to me

PR #1928, opened 2026-08-26
Don't add nasa bench recipe sometimes (VaraibleHorizons compat)

LazyFlesh:
## Summary

Don't add Nasa bench recipe when VariableHorizons is loaded and NoRocket is active.
This is alternative to doing a recipe removal when pack is done loading. 

## Checklist


EnderProyects:
I think the variable should dissable the rocket recipes, not the nasa workbench recipe itself, it shouldnt but now that @S4mpsa is doing the eoh rework maybe it uses it in the controller of the voidcraft assembler xd

LazyFlesh:
> I think the variable should dissable the rocket recipes, not the nasa workbench recipe itself, it shouldnt but now that @S4mpsa is doing the eoh rework maybe it uses it in the controller of the voidcraft assembler xd

...im not sure where the rocket recipes are, but ill take a look.

LazyFlesh:
Found em. Those we can delete without looking through the whole recipe registry.

...In such a case, I suppose this isn't required.

PR #1837, opened 2026-06-29
Add missing ore variants to the space shuttle craft

boubou19:
close https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/24927
depends on https://github.com/GTNewHorizons/GT5-Unofficial/pull/7307

before:
image

after:
image

PR #1933, opened 2026-08-28
Improve planet recipes with naturally generated ores

boubou19:
## Summary

Because of the new ore variants, some recipes were lacking ore variants in their recipe. It's mostly a followup PR to fix the same kind of things as https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/1837, this time for the NEI Planets recipes.

Depend on https://github.com/GTNewHorizons/GT5-Unofficial/pull/7886 to be merged and tagged to pass CI.

## Checklist
- [x] This PR requires another PR in order to merge


src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
```diff
@@ -8882,7 +8871,7 @@ private void makeNEIPlanetRecipes() {
 
             // Technically T10
             // Deep Dark
-            for (OrePrefixes orePrefix : allOrePrefixes) {
+            for (OrePrefixes orePrefix : getOrePrefixesVariants(Materials.Rubidium)) {
```
UltraProdigy: Rubidium has no OreMix so this won't return anything

UltraProdigy (changes requested):

src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
boubou19: I made a new method to add default ore prefixes, just like the amunra script does it to fix this.

src/main/java/com/dreammaster/gthandler/recipes/AssemblerRecipes.java
```diff
@@ -8382,20 +8383,35 @@ private void makeCoilRecipes() {
 
     }
 
+    private OrePrefixes[] getOresFromMaterial(Materials materials) {
+        List<OrePrefixes> variants = Arrays.asList(getOrePrefixesVariants(materials));
```
UltraProdigy: This returns a fixed sized list so variants.add will crash on startup, should be made mutable instead

UltraProdigy (commented):
Last thing and this is good to rock

UltraProdigy (approved):
🔥

PR #1952, opened 2026-09-12
Fix/futurum stripped logs to planks

vakus:
## Summary

I was playing on latest beta, and noticed that some stripped logs can't be converted to planks, this PR adds those recipes.
Issue related to this PR https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/25795

image
image
image

I omitted the recipe screenshots for all other woods, however all of stripped logs added by et-futurum-requiem for BiomesOPlenty (except Dead Log and Giant Flower Stem which do not have respective planks) and Witchery have received the same treatment.

## Checklist

### DIFF
--- a/src/main/java/com/dreammaster/gthandler/GT_CraftingRecipeLoader.java
+++ b/src/main/java/com/dreammaster/gthandler/GT_CraftingRecipeLoader.java
@@ -7,6 +7,7 @@ import static gregtech.api.enums.Materials.Bronze;
 import static gregtech.api.enums.Mods.AE2FluidCraft;
 import static gregtech.api.enums.Mods.AdventureBackpack;
 import static gregtech.api.enums.Mods.AppliedEnergistics2;
+import static gregtech.api.enums.Mods.BiomesOPlenty;
 import static gregtech.api.enums.Mods.BuildCraftFactory;
 import static gregtech.api.enums.Mods.Chisel;
 import static gregtech.api.enums.Mods.Computronics;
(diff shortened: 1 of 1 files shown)


src/main/java/com/dreammaster/gthandler/GT_CraftingRecipeLoader.java
```diff
@@ -2443,6 +2445,100 @@ public void run() {
...
+                GameRegistry.addShapelessRecipe(
+                        getModItem(BiomesOPlenty.ID, "planks", 4, 4),
+                        getModItem(EtFuturumRequiem.ID, "bop_log_stripped2", 1, 0));
+
+                // Stripped Magic
+                GameRegistry.addShapelessRecipe(
+                        getModItem(BiomesOPlenty.ID, "planks", 4, 5),
+                        getModItem(EtFuturumRequiem.ID, "bop_log_stripped2", 1, 1));
+
+                // Stripped Magic
```
UltraProdigy: Copy / paste error I'm assuming

UltraProdigy (changes requested):
This is a good change, this should also probably include the bark and the wood_stripped. Also maybe not in this PR but this also cause me to notice several of the log -> wood recipes don't actually exist yet.

src/main/java/com/dreammaster/gthandler/GT_CraftingRecipeLoader.java
vakus: I removed duplicated code, re-tested it in latest daily, and all added recipes are there as expected

vakus:
> This is a good change, this should also probably include the bark and the wood_stripped. Also maybe not in this PR but this also cause me to notice several of the log -> wood recipes don't actually exist yet.

I added the recipes for bark and stripped woods as well now

PR #1953, opened 2026-09-12
Fix HazardousItems profiler interval check

Chitak985:
## Summary
The profiler call check should run every second but the condition is backwards, so it actually constantly runs and stops for 1 ms every second. This should fix it and make it run once and wait a full second before doing so again.

## Checklist


UltraProdigy (approved):
Ty!

PR #1957, opened 2026-09-13
Add the Primitive circuit line to NAC

chrombread:
## Summary

This circuit line was planned from when NAC was first put into dailies, but we put it off so that we could ship NAC faster.

This whole chain is pending textures. When those are made i will show in-game images but here is the gist of it:

Crystal board -> bio coordinator -> Unattuned Circuitry
Unattuned Circuitry + optical SMDS + cosmic neut + (maybe something else -> matrix 25s UEV -> Primed Unattuned Circuitry
Primed Unattuned Circuitry + ASOC + respective bolts + respective wires -> matrix 25s UHV -> ULV/LV/MV/HV/EV at (64/32/16/8/4) per circuitry.

The Primitive calibration bonuses will also be really wacky as well: granting bonus circuits, insane random speedups, and possibly a new fluid to be used elsewhere in the NAC proccess.

## TODO
- [x] finalize recipes

## Checklist
- [X] I have tested this PR in DevEnv
- [X] This PR is in compliance with the [GTNH AI Policy](https://github.com/GTNewHorizons/GTNH-Dev-Doc/blob/master/AI_POLICY.md)
- [X] This PR requires another PR in order to merge https://github.com/GTNewHorizons/GT5-Unofficial/pull/8049#issue-5442744016


fehling135:
Maybe too many optical SMD? 
Take HV circuits as an example, each HV circuit takes 3*2 advanced SMD in CAL, but here it takes 5*2 optical
Moreover, SoC recipes do not take SMDs in CAL

chrombread:
> Maybe too many optical SMD? Take HV circuits as an example, each HV circuit takes 3_2 advanced SMD in CAL, but here it takes 5_2 optical Moreover, SoC recipes do not take SMDs in CAL

what i was going for with these recipes was the board becomes a living organism prepared to adapt into any of these circuit types in the first matrix step, and then it gets attuned to the specific circuit type with the solder/components of that circuit type. hopefully this will be conveyed better with textures.
tldr the board is its own type of circuit that melds into the other circuit types

~~as for the cost in smds, wouldnt the number be 80 optical smds : 1 primed unattuned circuitry : 8 hv circuits?. the ratio is 10 optical smds : 1 hv circuit in that case. which seems pretty reasonable, if not a little generous.~~ taking another your message had the right rates but github formatted it wrong.

I thought that 10 smds was a good enough tradeoff but looking at how higher tier circuits are treated this could be tuned down a bit. i think it may need to become 4 instead of 16 to match better. thanks for catching that

EnderProyects:
> possibly a new fluid to be used elsewhere in the NAC proccess.

Just saying, end line idea is to add alternate recipes that add buffs to already existing recipes, WIP molten encased translokite could work in here

chrombread:
> > possibly a new fluid to be used elsewhere in the NAC proccess.
> 
> Just saying, end line idea is to add alternate recipes that add buffs to already existing recipes, WIP molten encased translokite could work in here

we ended up scrapping the fluid idea, but if something else  can exist that accomplishes the same effect that isn't sourced from the NAC, it would be fine.

chrombread:
undrafting, ready for review. works on our server after the accompanying PR is merged.

serenibyss (approved):

boubou19 (approved):

PR #1958, opened 2026-09-14
remove duplicated ardite & manyullyn saw blades

danyadev:
## Summary

GT already generates Ardite and Manyullyn saw blades, no need to make our own.

Postea transformer tested, and there's no recipe regression

| Before | After |
| - | - |
| image | image |
| image | image |

Needed for https://github.com/GTNewHorizons/GT5-Unofficial/pull/7998

## Checklist


Nikolay-Sitnikov:
If the two items are supposed to act identically to each other, then we should just use Postea to merge one into the other. If they're different (can be obtained or used in different ways), then they should be differentiated by giving them different names.

danyadev:
yea, they turned out to be duplicated items. Removed them

src/main/java/com/dreammaster/main/MainRegistry.java
```diff
@@ -393,6 +394,8 @@ public void PostLoad(FMLPostInitializationEvent PostEvent) {
         if (!Loader.isModLoaded(BPPConverter.BPP_MOD_ID)) BPPConverter.doPostInitialization();
 
         if (IndustrialCraft2.isModLoaded()) IC2Converter.doPostInitialization();
+
+        if (IguanaTweaksTinkerConstruct.isModLoaded()) IguanaConverter.doPostInitialization();
```
Nikolay-Sitnikov: Does GT5U not generate those items if Iguana Tinker Tweaks isn't present? I'd think that those items always exist as long as GT5U exists, in which case this should run unconditionally.

Nikolay-Sitnikov (approved):
One thing I need you to confirm, after which you can merge.

Dream-Master:
@danyadev

src/main/java/com/dreammaster/main/MainRegistry.java
danyadev: the generation is independent from iguana presence, removed the check

PR #1963, opened 2026-09-18
Fix UIV Solenoids not using any coolant

Auynonymous:
## Summary
For some reason, UIV solenoids did not use any coolant whatsoever, and instead used 32 ingots of Transmetal
This logic makes no sense when compared to previous solenoids, or even to the umv solenoid
(UMV does use spacetime, but technically it is a coolant)

This pr changes it to use 1024K L (4x uev, following previous tier logic) of supercoolant instead, and should be changed to a new coolant if and when it is added.

This does make them cheaper, and removes ~80k Transmetal from Galli Stabs. This was a nonsense cost, so it should be changed regardless.

breaking change since this changes the recipe

## Checklist


Nockyx (approved):

Dream-Master:
@boubou19

boubou19 (approved):

